PR #72, opened 2015-04-10
Convert to new netflixoss build

rspieldenner:
You may want to consider moving dependencies into a build.gradle in the subproject and making sure all the subprojects actually need all the dependencies in the subprojects block.


jcacciatore:
Manually merged

PR #94, opened 2015-12-08
HSCAN, ZSCAN and SSCAN added

ipapapa:
Added support for SCAN operations inside a single key.


dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
```diff
@@ -533,6 +533,42 @@ public Long execute(final Jedis client, final ConnectionContext state) throws Dy
         });
     }
 
+    
+    @Override
+    public ScanResult<Map.Entry<String, String>> hscan(final String key, final int cursor) {
```
jcacciatore: ScanResult<T> is Jedis class (not interface). I'd prefer not to paint ourselves into this corner again, as we did with Response<T>. Client libraries that leak their underlying dependencies but aren't strictly delegating as passthrough quickly get into trouble...

On the other hand, I can appreciate that the Dyno client should be a drop-in replacement (to the extent this makes sense) for Jedis in order for client applications to seamlessly switch between the two. This approach does make Dyno client code more convoluted however, and will tie the Dyno client to specific version(s) of the Jedis client.

Thoughts ?

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
```diff
@@ -533,6 +533,42 @@ public Long execute(final Jedis client, final ConnectionContext state) throws Dy
...
+    	return d_hscan(key, cursor).getResult();
+    }
+    
+    public OperationResult<ScanResult<Map.Entry<String, String>>> d_hscan(final String key, final int cursor){
+    	
+    	return connPool.executeWithFailover(new BaseKeyOperation<ScanResult<Map.Entry<String, String>>>(key, OpName.HSCAN) {
```
jcacciatore: Hash operations now support compression. The values in the cursor returned for hscan, sscan, and zscan all need to support this now.

ipapapa: Good point, I will add compression on HSCAN.

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
ipapapa: I agree with you; IMHO locking the Dyno client to a specific version of the Jedis Client is not bad. The current path of Jedis and Redis seem to be to focus on the Redis Cluster. Hence, as long as Dyno provides extra features to the client, I do not see anybody switching to the vanilla Jedis client from Dyno. Once a new client API is introduced for single node Redis and that API call is widely requested, we can revisit this. Hence, shall we extend ScanResult instead of overriding it?

jcacciatore: Good points.

I would prefer composition over inheritance for this case. Rather than extending ScanResult, we would create a parameterized interface in dyno-core as well as an implementation class. The implementation class would contain a private reference to ScanResult.

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
jcacciatore: You're correct - hscan is the only operation that needs it (not sscan or zscan). Thanks!

PR #101, opened 2016-03-16
Update gradle to 2.6

ipapapa:


jcacciatore:
Thank you @ipapapa for doing this!

PR #105, opened 2016-04-13
Bug Fix - host pool can run out of connections

jcacciatore:
**Detailed Description**
The 1.4.0 release discards connections that encounter socket exceptions due to an insidious underlying buffer issue. If the dynomite connection is severed while all connections are in use all of them will be discarded and the pool will not replenish itself immediately. The result is that all requests will fail-over to a secondary AZ. The fail-over is by design and the correct behavior and if the HostSupplier notifies the client that dynomite is Down, all is well. If the network connection simply had an issue and Dynomite is actually UP, then the HostSupplier will not notify the client and the pool will never replenish its connections to that host.

This is particularly problematic if the client application has wrapped the dyno client call in a load-shedding layer such as Hystrix and left the default timeouts in place (Hystrix is 1 second, Dyno client timeout default is 2 seconds) therefore since the client app timeout fires first the fallback for dyno client does not occur and all requests fail.

**Solution**
Detect when the pool is exhausted and notify the health tracker which will immediately put the pool into a reconnect state after which, all of the client's current reconnect + failover logic will apply.

### DIFF
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/exception/PoolExhaustedException.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/exception/PoolExhaustedException.java
@@ -24,11 +24,11 @@ public class PoolExhaustedException extends DynoConnectException {
 
     public PoolExhaustedException(HostConnectionPool hostConnectionPool, Throwable t) {
         super(t);
-        hcp = hostConnectionPool;
+        this.hcp = hostConnectionPool;
     }
 
     public HostConnectionPool getHostConnectionPool() {
-        return hcp;
+        return this.hcp;
     }
 }
 
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/exception/PoolExhaustedException.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/exception/PoolExhaustedException.java
@@ -1,22 +1,34 @@
 package com.netflix.dyno.connectionpool.exception;
 
-public class PoolExhaustedException extends DynoConnectException { 
+import com.netflix.dyno.connectionpool.HostConnectionPool;
 
-	/**
-	 * 
-	 */
+/**
+ * Indicates there are no connections left in the host connection pool.
+ */
+public class PoolExhaustedException extends DynoConnectException {
+
+    private final HostConnectionPool hcp;
+
 	private static final long serialVersionUID = 9081993527008721028L;
 
-	public PoolExhaustedException(String message) {
+
+	public PoolExhaustedException(HostConnectionPool hostConnectionPool, String message) {
 		super(message);
+        this.hcp = hostConnectionPool;
 	}
 
-	public PoolExhaustedException(Throwable t) {
-		super(t);
-	}
+    public PoolExhaustedException(HostConnectionPool hostConnectionPool, String message, Throwable cause) {
+        super(message, cause);
+        this.hcp = hostConnectionPool;
+    }
 
-	public PoolExhaustedException(String message, Throwable cause) {
-		super(message, cause);
-	}
+    public PoolExhaustedException(HostConnectionPool hostConnectionPool, Throwable t) {
+        super(t);
+        hcp = hostConnectionPool;
+    }
+
+    public HostConnectionPool getHostConnectionPool() {
+        return hcp;
+    }
 }
 
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostConnectionPoolImpl.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostConnectionPoolImpl.java
@@ -23,6 +23,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 
+import com.netflix.dyno.connectionpool.exception.PoolExhaustedException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -108,6 +109,9 @@ public class HostConnectionPoolImpl<CL> implements HostConnectionPool<CL> {
 
 	@Override
 	public void markAsDown(DynoException reason) {
+        if (Logger.isDebugEnabled()) {
+            Logger.debug(String.format("Marking Host Connection Pool %s DOWN", getHost()));
+        }
 		
 		ConnectionPoolState<CL> currentState = cpState.get();
 		
@@ -304,7 +308,7 @@ public class HostConnectionPoolImpl<CL> implements HostConnectionPool<CL> {
                     }
 				}
 				monitor.incConnectionCreateFailed(host, e);
-				throw new DynoConnectException(e);
+				throw new DynoConnectException(e); // TODO - FatalConnectionException ?
 			}
 		}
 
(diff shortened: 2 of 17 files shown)


dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallback.java
```diff
@@ -120,9 +120,14 @@ public HostSelectionWithFallback(ConnectionPoolConfiguration config, ConnectionP
 			try { 
 				return hostPool.borrowConnection(duration, unit);
 			} catch (DynoConnectException e) {
-				lastEx = e;
-				cpMonitor.incOperationFailure(null, e);
-				useFallback = true;
+				if (e instanceof PoolExhaustedException) {
```
timiblossom: We got PoolExhaustedException throwing when the pool has zero connections, if we don't do fall back immediately, when are we going to do the fall back?  Also, what happens if the pool can actually never have any connections such as in RF=2? In this RF=2 case, if we don't fall back immediately, there would be a huge performance degradation, no?

dyno-core/src/main/java/com/netflix/dyno/connectionpool/exception/PoolExhaustedException.java
```diff
@@ -1,22 +1,34 @@
...
+        super(message, cause);
+        this.hcp = hostConnectionPool;
+    }
 
-	public PoolExhaustedException(String message, Throwable cause) {
-		super(message, cause);
-	}
+    public PoolExhaustedException(HostConnectionPool hostConnectionPool, Throwable t) {
+        super(t);
+        hcp = hostConnectionPool;
```
ipapapa: hcp ---> this.hcp

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostConnectionPoolImpl.java
```diff
@@ -304,7 +308,7 @@ private ConnectionPoolActive(HostConnectionPoolImpl<CL> cp) {
                     }
 				}
 				monitor.incConnectionCreateFailed(host, e);
-				throw new DynoConnectException(e);
+				throw new DynoConnectException(e); // TODO - FatalConnectionException ?
```
ipapapa: I am assuming this is just to make the exceptions cleaner.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/exception/PoolExhaustedException.java
jcacciatore: Thanks - consistency is good :-)

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostConnectionPoolImpl.java
jcacciatore: Good catch - I was thinking through solutions and put that as a marker for one of them. I forgot to remove it.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostConnectionPoolImpl.java
```diff
@@ -359,15 +370,15 @@ public boolean closeConnection(Connection<CL> connection) {
 				conn = availableConnections.poll(duration, unit);
 			} catch (InterruptedException e) {
 				Logger.info("Thread interrupted when waiting on connections");
-				throw new DynoConnectException(e);
+				throw new DynoConnectException(e); // TODO Change this to DynoBorrowedConnectionException(e)
```
jcacciatore: Decided against the comment; solution doesn't require a specific exception for this but uses PoolExhausted exception instead.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallback.java
jcacciatore: Currently in this scenario the fall back does not occur immediately, it occurs after the timeout expires when polling the queue. By default this is 2 seconds, which is a long time to wait for a fall back. This will cause huge performance degradation (and did which is why the Hystrix timeout kicked in at the app layer).

For RF=3 the solution is good. Even in the case of a zone outage, the client will fail-fast and fall back to the other zones.

For the RF=2 case however, this is a good catch! The performance won't be degraded but the health tracker will infinitely be searching for a Dynomite node in the zone, which is not desirable. I will add the RF=2 client changes into this PR to account for this case.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallback.java
```diff
@@ -328,8 +333,29 @@ public void initWithHosts(Map<Host, HostConnectionPool<CL>> hPools) {
...
+            }
+        }
+
+        Set<Integer> uniqueCounts = new HashSet<>(groups.values());
+
+        if (uniqueCounts.size() > 1) {
+            throw new RuntimeException("Invalid configuration - replication factor cannot be asymmetric");
+        }
+
+        return uniqueCounts.toArray(new Integer[uniqueCounts.size()])[0];
```
timiblossom: What happens that during an initialization, the 1st entry has fewer than the desired elements or count.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallback.java
```diff
@@ -120,7 +123,7 @@ public HostSelectionWithFallback(ConnectionPoolConfiguration config, ConnectionP
 			try { 
 				return hostPool.borrowConnection(duration, unit);
 			} catch (DynoConnectException e) {
-				if (e instanceof PoolExhaustedException) {
+				if (e instanceof PoolExhaustedException && replicationFactor.get() == 3) {
```
timiblossom: Why is it 3 here?  What happens if RF = 2 or 5?

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/MonitorConsole.java
```diff
@@ -159,6 +159,7 @@ public TokenPoolTopology getTopology(String cpName) {
             config.put("socketTimeout", String.valueOf(cpConfig.getSocketTimeout()));
             config.put("timingCountersResetFrequencyInSecs",
                     String.valueOf(cpConfig.getTimingCountersResetFrequencySeconds()));
+            config.put("replicationFactor", String.valueOf(pool.getTopology().getReplicationFactor()));
```
timiblossom: This new property is only used in one place and at the same time, I also see the code calculates the RF dynamically.  Would there be a wrong logic later on in setting this as well as automatic calculation?

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallback.java
jcacciatore: After reviewing again, I realized that bit of code actually introduces a logic issue. I'm going to remove the replication check and add a comment for clarification.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallback.java
jcacciatore: @timiblossom spoke offline; this is not an issue

PR #106, opened 2016-04-13
Track connection pools via the MonitorConsole.

jcacciatore:
**Description**
Since the Dyno client is not a singleton (by design) there have been instances where client applications were instantiating a Dyno client per application thread but did not realize they were doing so. The code in the monitor console uses the connection pool name and relies on it to be unique.

**Resolution**
Ensure the name is unique and if not, make it so by appending the current system time to the name. The names can be discovered using `MonitorConsole.getMonitorNames()`. Note this does not affect metrics reporting or anything else; the scope of this is confined to the Monitor Console only.


ipapapa:
Please rebase.

PR #109, opened 2016-05-04
Binary ops

ipapapa:
Support for binary operation


jcacciatore:
Thanks for doing this Yannis. As we discussed, we can do a more proper refactoring to separate the interfaces when we expose more of the commands.

PR #110, opened 2016-05-18
RF=2 metrics bug fix

jcacciatore:
When the server topology has replication factor of 2 AND there are dyno client instances that are located in a rack that has no corresponding dynomite hosts, the failover metrics are not accurate since every request is viewed as failing over to a remote rack.

For example, if dynomite hosts are in us-east-1c and us-east-1d but dyno client instances are in 1c, 1d, **and also 1e** all requests from dyno client instances in 1e would be recorded as failover requests.

This PR fixes this issue.

### DIFF
--- a/dyno-core/src/test/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallbackTest.java
+++ b/dyno-core/src/test/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallbackTest.java
@@ -326,6 +326,37 @@ public class HostSelectionWithFallbackTest {
 		verifyExactly(hostnames, "h5", "h6");
 	}
 
+    @Test
+    public void testReplicationFactorOf3WithDupes() {
+        cpConfig.setLoadBalancingStrategy(LoadBalancingStrategy.TokenAware);
+        cpConfig.withTokenSupplier(getTokenMapSupplier());
+
+        HostSelectionWithFallback<Integer> selection = new HostSelectionWithFallback<Integer>(cpConfig, cpMonitor);
+
+        List<HostToken> hostTokens = Arrays.asList(
+                new HostToken(1383429731L, new Host("host-1", -1)), // Use -1 otherwise the port is opened which works
+                new HostToken(2815085496L, new Host("host-2", -1)), // but slows down the test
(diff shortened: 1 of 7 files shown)


dyno-core/src/test/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallbackTest.java
```diff
@@ -326,6 +326,37 @@ public void testGetConnectionsFromRingWhenHostDown() throws Exception {
 		verifyExactly(hostnames, "h5", "h6");
 	}
 
+    @Test
+    public void testReplicationFactorOf3WithDupes() {
```
ipapapa: What does Dupes mean? Do we need a Test for RF2?

ipapapa:
Based on my understanding of the implementation (which may be totally off), if there is no available node in the local rack, we assume that we are using RF=2. This may create some corner cases, so I was wondering whether we can cross-validate whether the Dynomite server is RF=2 through cluster_describe.

dyno-core/src/test/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallbackTest.java
jcacciatore: There is a test for RF2, [here](https://github.com/jcacciatore/dyno/blob/3b482d9d0ed41a4ccd6bb595dc79edc779686d21/dyno-core/src/test/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallbackTest.java#L382).

The client contacts every dynomite server in determining the topology therefore it sees a lot of duplicate info.

jcacciatore:
@ipapapa The logic to determine the replication factor is based on the cluster_describe from the side-car (currently). Where do you see an assumption of RF2 ?

dyno-core/src/test/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallbackTest.java
ipapapa: ha, I See it in the "new Host". Thanks covered.

PR #118, opened 2016-06-02
If the local zone is not known and the client is configured to be tok…

jcacciatore:
…en aware an NPE is thrown. This fixes that issue.


ipapapa:
Looks good to me.

PR #122, opened 2016-07-29
Dual-Write functionality

jcacciatore:
Also fixed a host status tracker bug that kept inactive hosts around indefinitely even though the HostSupplier has stopped reporting them.


dyno-core/src/main/java/com/netflix/dyno/connectionpool/ConnectionPool.java
```diff
@@ -148,4 +148,6 @@
      */
     HealthTracker<CL> getHealthTracker();
 
+    boolean isIdle();
```
ipapapa: isIdle applicable to both the shadow and the main pool for the primary cluster? IF it is only for the shadow, can we rename it to isIdleShadow()?

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/ConnectionPoolConfigurationImpl.java
```diff
@@ -87,6 +87,29 @@ public ConnectionPoolConfigurationImpl(String name) {
 		this.localRack = ConfigUtils.getLocalZone();
 		this.localDataCenter = ConfigUtils.getDataCenter();
 	}
+
+	public ConnectionPoolConfigurationImpl(ConnectionPoolConfigurationImpl config) {
```
ipapapa: How do we need another constructor? Can you add a few lines of comments above?

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoDualWriterClient.java
```diff
@@ -0,0 +1,188 @@
...
+import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Client that provides 'dual-write' functionality. This is useful when clients wish to move from one dynomite
+ * cluster to another, for example to upgrade cluster capacity.
```
ipapapa: This is a nice comment, we can reference to this line when people ask us how to scale up a cluster.

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
```diff
@@ -3255,8 +3269,74 @@ public Builder withDiscoveryClient(DiscoveryClient client) {
             return this;
         }
 
+        public Builder withShadowClusterName(String shadowCluster) {
+            shadowClusterName = shadowCluster;
+            return this;
+        }
+
+        public Builder withShadowHostSupplier(HostSupplier shadowHostSupplier) {
```
ipapapa: Do we need two Host Suppliers? One for the primary and one for the shadow? or is it usually the case that we will have just one host supplier?

dyno-core/src/main/java/com/netflix/dyno/connectionpool/ConnectionPool.java
jcacciatore: That method is on the ConnectionPool interface so it doesn't make sense to name it specifically for a shadow implementor.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/ConnectionPoolConfigurationImpl.java
jcacciatore: The shadow pool needs its own configuration object. Rather than making ConnectionPoolConfigurationImpl ```Cloneable``` I added a copy constructor. 

I will add a javadoc comment to the constructor - thanks

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoDualWriterClient.java
jcacciatore: Thanks - I am writing up a wiki page now as well

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
jcacciatore: Good question! We actually do need two HostSupplier objects b/c there is a 1:1 relationship with a HostSupplier and the application they are retrieving host information for. Note however that there is only a single DiscoveryClient object that gets shared between both HostSupplier objects.

PR #123, opened 2016-08-10
Dualwriteapi

ipapapa:
* Adding missing Apache Licenses so that building does not complain anymore;
* Replaced deprecated `blpop` and `brpop` commands with newer implementations from Jedis 2.8;
* Adding support for dual-write Jedis commands;
* Adding support for  binary`set` and `setex` in dual-write;
* Further minor fixes.

### DIFF
--- a/dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoDualWriterClient.java
+++ b/dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoDualWriterClient.java
@@ -1,3 +1,18 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.netflix.dyno.jedis;
 
 import com.netflix.dyno.connectionpool.ConnectionPool;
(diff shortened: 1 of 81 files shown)


dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoDualWriterClient.java
```diff
@@ -84,14 +122,23 @@ private boolean sendShadowRequest(String key) {
...
     public interface Dial {
         /**
          * Returns true if the given value is in range, false otherwise
          */
         boolean isInRange(String key);
 
-        void setRange(int range);
+        boolean isInRange(byte[] key);
+
+		void setRange(int range);
```
shailesh33: indentation ... may be in the next PR

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoDualWriterClient.java
```diff
@@ -191,5 +244,535 @@ public String setex(final String key, int seconds, String value) {
         return targetClient.setex(key, seconds, value);
     }
 
+    @Override
+    public Long del(final String key) {
```
shailesh33: This is applicable to all operations being implemented here. Both del and d_del are public and we wont get the dual write functionality if someone calls d_ variants for an operation.
Correct me if I am wrong, but you should call d_ variant of the dual writer here. And that guy should implement the dual writer.

jcacciatore:
LGTM

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoDualWriterClient.java
ipapapa: fixed

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoDualWriterClient.java
ipapapa: My understanding is that d_ variants are representing the result of executing an operation, aka `OperationResult`. So a client wants to use a similar to Jedis interface that leverages String, Long, Boolean etc.

PR #124, opened 2016-08-11
Using the private IP address if it can find it, otherwise the hostname.

ipapapa:
Creating new `Host` constructors that can take the private IP address. If private IP address is provided by the Discovery service (Eureka), then the client will use that. If a private IP address is not found then it will use the hostname.

### DIFF
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
@@ -24,7 +24,8 @@ import java.net.InetSocketAddress;
  */
 public class Host {
 
-	private final String name;
+	private final String hostname;
+	private final String ipAddress;
 	private int port;
 	private Status status = Status.Down;
 	private InetSocketAddress socketAddress = null;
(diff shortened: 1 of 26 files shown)


dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
```diff
@@ -35,27 +36,57 @@
...
-	public Host(String name, int port) {
-		this(name, port, Status.Down);
+	public Host(String hostname, int port) {
+		this(hostname, null, port, Status.Down);
 	}
 	
-	public Host(String name, Status status) {
-		this.name = name;
+	public Host(String hostname, Status status) {
+		this.hostname = hostname;
```
shailesh33: you have one superset constructor which takes all possible arguments. You should redirect all the other constructors to that one like you do on line 39.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
```diff
@@ -35,27 +36,57 @@
...
+	
+	public Host(String hostname, String ipAddress, Status status) {
+		this.hostname = hostname;
+		this.ipAddress = ipAddress;
+		this.port = -1;
+		this.status = status;
+	}
+
+	public Host(String name, String ipAddress, int port, Status status) {
+		this.hostname = name;
```
shailesh33: code repetition from line 51

ipapapa: a single constructor as proposed above will cover that.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/ConnectionPoolImpl.java
```diff
@@ -197,10 +197,10 @@ public boolean removeHost(Host host) {
 			cpHealthTracker.removeHost(host);
 			cpMonitor.hostRemoved(host);
 			hostPool.shutdown();
-            Logger.info(String.format("Remove host: Successfully removed host %s from connection pool", host.getHostName()));
+            Logger.info(String.format("Remove host: Successfully removed host %s from connection pool", host.getHostAddress()));
```
shailesh33: are you ok with IP addresss getting printed everwhere?

ipapapa: Either the hostname or the private IP address will be printed. Based on what has been used the correct one will be printed. I am not sure I am following your question.

shailesh33: Agreed. I just want to make sure that printing private ip address (which will be the case from now on) will not cause issues while debugging.

PR #126, opened 2016-08-13
Bug fixes for dual write functionality

jcacciatore:
* Property based settings now work properly
* Fixed bug with injecting HostSupplier
* Now support dual write functionality if client app is using d_ methods in DynoJedisClient

### DIFF
--- a/dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoDualWriterClient.java
+++ b/dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoDualWriterClient.java
@@ -174,74 +174,106 @@ public class DynoDualWriterClient extends DynoJedisClient {
 
     @Override
     public Long append(final String key, final String value) {
+        return this.d_append(key, value).getResult();
+    }
+
+    @Override
+    public OperationResult<Long> d_append(final String key, final String value) {
         writeAsync(key, new Callable<OperationResult<Long>>() {
             @Override
             public OperationResult<Long> call() throws Exception {
-                return d_append(key, value);
+                return DynoDualWriterClient.super.d_append(key, value);
             }
         });
 
-        return targetClient.append(key, value);
+        return targetClient.d_append(key, value);
     }
 
     @Override
     public String hmset(final String key, final Map<String, String> hash) {
+        return this.d_hmset(key, hash).getResult();
+    }
+
+    @Override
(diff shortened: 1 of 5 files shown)


dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoDualWriterClient.java
```diff
@@ -174,74 +174,106 @@ public void setRange(int range) {
...
+        return this.d_append(key, value).getResult();
+    }
+
+    @Override
+    public OperationResult<Long> d_append(final String key, final String value) {
         writeAsync(key, new Callable<OperationResult<Long>>() {
             @Override
             public OperationResult<Long> call() throws Exception {
-                return d_append(key, value);
+                return DynoDualWriterClient.super.d_append(key, value);
```
ipapapa: Sounds weird that we are calling DynoDualWriterClient instead DynoDualWriterClient class. Wouldn't `this` not work here?

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoDualWriterClient.java
```diff
@@ -174,74 +174,106 @@ public void setRange(int range) {
 
     @Override
     public Long append(final String key, final String value) {
+        return this.d_append(key, value).getResult();
+    }
+
+    @Override
+    public OperationResult<Long> d_append(final String key, final String value) {
```
ipapapa: was implementing the `d_xxx` needed for the dual write fix, or they were done as part of our conversation?

ipapapa:
Other than my comments, logically LGTM.

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoDualWriterClient.java
jcacciatore: Using `this` will cause an infinite loop. Note that I have to specify `DynoDualWriterClient` in order to call **`super`**

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoDualWriterClient.java
jcacciatore: It was not necessary for the fix. I either needed to support it or modify pappy code calling these methods.

ipapapa:
Merging, tagging and baking to test.

Issue #50, opened 2014-11-12
Connection pool configuration contains port and ignores supplied port from host


cporter71:
It is not clear if this is misunderstanding on my end or if this is the intended behavior but I see that your connection pool configuration includes a host port.  That port is then used to overwrite the port in a supplied host:

ConnectionPoolImpl.java
    public boolean addHost(Host host, boolean refreshLoadBalancer) {
        host.setPort(cpConfiguration.getPort());

In our deployed env we would have N-redis instances on a single host.  Following the documentation we would then setup N nodes (dynomite + redis) on that host.  This would mean N ports.

Please let me know what I am missing.

cporter71:
Thanks opuneet -

In terms of modeling I would think that you DO NOT want the port in the connection pool configuration but rather with the host.  Would you add the host name to the connection pool configuration?  No, since then you would have to create N (host count) ConnectionPoolConfiguration objects.  Adding port to the ConnectionPoolConfiguration forces a similar issue for those use cases where a single host runs more than a single instance of redis/dynomite.  

From a modeling perspective "port" relates to the host, not the runtime behavior of connection management.  It seems that using port from the Host object works for the Netflix use case as well as providing flexibility to those uses cases outside of 1-to-1 dynomite-node-to-host model and for that reason feels like the right direction.

alosix:
I've recently run into this as well. I've mostly gotten around it by subclassing the host object to ignore the port set done by the connection pool.

For me, instead of a bunch of smaller AWS instances we use some rather large boxes (128gb/48 core).  Redis and dyno seem to work better with many instances per box. Its easier to configure this with different ports than it is to beg for more ips on the lan for this.

So I'd most definitely vote to keep the port configuration at the host level.  The dynomite server instance doesn't seem to have any issues with this.

ipapapa:
@alosix What are the advantages you have seen with running multiple Dynomite/Redis in the same bigger instance? If the instance goes down, you need to warm up two token ranges.

alosix:
@ipapapa  For me its to more efficiently use the hardware at hand for processing higher loads vs availability. This is being using in some lambda counting in conjunction with twitter's summingbird, So the data this creates is ephemeral and used for about 2 hours until the map reduce side dumps the finalized counts into cassandra for longer term storage.

ipapapa:
@alosix sounds a very interesting use case. Thank you for the information.

PR #127, opened 2016-08-15
Issue #50

alosix:
Initial pull request to get some input on these changes.  I'll get the build.gradle backed out shortly, those changes were handy to get them into my local repo for an sbt project.

Changes:
* The host object is now mostly immutable, allowing only the status to be changed and I'm not 100% sure its used so we might be able to remove that.
* The host object's equals, hashcode, and comparable take hostName, port, and rack into consideration.
* Ports can only be set on the host objects in their constructors. 
* The HttpTokenSupplier has an option to supply a default port if they are not specified in the json for compatibility with configurations that don't have ports and all use the same port as it was originally done.

* This patch will take a few minor code changes for clients that take them, but not much different than the 1.4.5 to 1.4.6 changes from what I've seen.

* Added unit tests to check selection with different ports as well as with the current way of doing things. All tests passing, though on of the rate limit ones is a little flakey.

* Also running in one our production environment and seems to be doing well with 3x servers with 10x dynomite and 10x redis running on each. Load seems evenly spread among all of the instances.

* Tried to finish up some of the renaming work that seems to be in progress for zone vs rack vs dc.

* I need to find the code style that's supposed to get used in this repo. I'll get these files formatted like that shortly. The code all in is kind of all over the place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Dyno
+# Dyno [![Build Status](https://secure.travis-ci.org/Netflix/dyno.png)](http://travis-ci.org/Netflix/dyno)
  
  Dyno encapsulates features necessary to scale a client application utilizing [Dynomite](https://github.com/Netflix/dynomite).
  
--- a/dyno-contrib/src/main/java/com/netflix/dyno/contrib/ArchaiusConnectionPoolConfiguration.java
+++ b/dyno-contrib/src/main/java/com/netflix/dyno/contrib/ArchaiusConnectionPoolConfiguration.java
@@ -63,7 +63,6 @@ public class ArchaiusConnectionPoolConfiguration extends ConnectionPoolConfigura
 		
 		String propertyPrefix = DynoPrefix + name; 
 		
-		//port = DynamicPropertyFactory.getInstance().getIntProperty(propertyPrefix + ".connection.port", super.getPort());
 		maxConnsPerHost = DynamicPropertyFactory.getInstance().getIntProperty(propertyPrefix + ".connection.maxConnsPerHost", super.getMaxConnsPerHost());
 		maxTimeoutWhenExhausted = DynamicPropertyFactory.getInstance().getIntProperty(propertyPrefix + ".connection.maxTimeoutWhenExhausted", super.getMaxTimeoutWhenExhausted());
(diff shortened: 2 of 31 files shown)


README.md
```diff
@@ -1,4 +1,4 @@
-# Dyno [![Build Status](https://secure.travis-ci.org/Netflix/dyno.png)](http://travis-ci.org/Netflix/dyno)
+# Dyno
```
jcacciatore: We don't want to remove the build badge

alosix: Yeah.. working on getting that back in, our guy doing the in house builds happened to grab the branch I was pull requesting from.

dyno-contrib/src/main/java/com/netflix/dyno/contrib/ArchaiusConnectionPoolConfiguration.java
```diff
@@ -36,7 +36,7 @@
 	
 	private static final String DynoPrefix = "dyno.";
 	
-	private final DynamicIntProperty port;
```
jcacciatore: I would favor removing rather than comment out all port changes

ipapapa:
@alosix we plan to test it out and provide further feedback very soon.

ipapapa:
@alosix I am so sorry for us to be so late; we had to release some important new features likes private IP address support and dual write support. I have built your changes and tested it. It looks good.

alosix:
@ipapapa Hi, sorry I just caught the message about getting the conflicts resolved. I'll take a few and get that done this week. 

thanks
Jason

ipapapa:
@alosix no problem, once we are done with your work, I will check out #137 so that it does not conflict you.

ipapapa:
@alosix any chance we can move forward with this PR and merge the code with the master branch? or there is no interest on it anymore?

alosix:
Still a good bit of interest, just been low on time. I'll make some today.

alosix:
Sorry if I'm a but fuzzy as its been a while since I worked on this.

It was possible (we were running it in production) Using a subclass of the
host class that included the port in the hash. The rest of things worked
well in that case, granted it may have broken other things that we didn't
hit in that use case.

PR #128, opened 2016-08-15
Write commands in dualwrite client

ipapapa:
* Local implementation of write commands in Dual writer;
* Adding `d_*` commands for the Dual writer;
* Removing some more deprecated Jedis commands.


jcacciatore:
LGTM

PR #130, opened 2016-08-26
Tokens for local Datacenter

ipapapa:
The PR includes the following:
* Limits Dyno to obtain and keep track of tokens only for nodes in each local datacenter. It addresses a security issue of making HTTP calls for Dynomite-manager tokens across regions. The only negative implication is when we use the JMX `getTopologySnapshot` we will only see same Datacenter tokens. Hence, we need to check client nodes in all regions to make sure the token distribution is correct.
* Fixes the bug that the HostToken received from the JSON does not contain a private IP address, hence a node could not be added in the Host connection pool.
* Added extra reliability in the way we perform requests for the `cluster_describe`. Before it would try one node and if it failed it would propagate the exception to the connection pool. Now it tries multiple nodes (twice by default) and propagates the last node to the connection pool. 
* Fixing some white spaces and some Java convention.

### DIFF
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
@@ -74,6 +74,11 @@ public class Host {
 		return hostname;
 	}
 	
+	public String getHostName() {
+	    return hostname;
+	}
+
+	
 	public String getIpAddress() {
 		return ipAddress;
 	}
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
@@ -31,6 +31,7 @@ public class Host {
 	private InetSocketAddress socketAddress = null;
 	
 	private String rack; 
+	private String datacenter;
 	
 	public static enum Status {
 		Up, Down;
@@ -97,7 +98,20 @@ public class Host {
 	
 	public Host setRack(String rack) {
 		this.rack = rack;
+        setDatacenter(rack);
 		return this;
+	}
+	
+	public String getDatacenter() {
+		return datacenter;
+	}
+	
+	/** 
+	 * Datacenter format us-east-x, us-west-x etc.
+	 * 
+	 */
+	private void setDatacenter(String rack) {
+		this.datacenter = rack.substring(0, rack.length() - 1);
 	}
 	
 	public Host setStatus(Status condition) {
@@ -142,6 +156,6 @@ public class Host {
 
 	@Override
 	public String toString() {
-		return "Host [hostname=" + hostname + ", ipAddress=" + ipAddress + ", port=" + port + ", rack: " + rack + ", status: " + status.name() + "]";
+		return "Host [hostname=" + hostname + ", ipAddress=" + ipAddress + ", port=" + port + ", rack: " + rack + ", datacenter: " + datacenter + ", status: " + status.name() + "]";
 	}
 }
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
@@ -17,150 +17,153 @@ package com.netflix.dyno.connectionpool;
 
 import java.net.InetSocketAddress;
 
+import com.netflix.dyno.connectionpool.impl.utils.ConfigUtils;
+
 /**
  * Class encapsulating information about a host.
+ * 
  * @author poberai
+ * @author ipapapanagiotou
  *
  */
 public class Host {
 
-	private final String hostname;
-	private final String ipAddress;
-	private int port;
-	private Status status = Status.Down;
-	private InetSocketAddress socketAddress = null;
-	
-	private String rack; 
-	private String datacenter;
-	
-	public static enum Status {
-		Up, Down;
-	}
-	
-	public Host(String hostname, int port) {
-		this(hostname, null, port, Status.Down);
-	}
-	
-	public Host(String hostname, Status status) {
-		this(hostname, null, -1, status);
-	}
-	
-	public Host(String hostname, int port, Status status) {
-		this(hostname, null, port, status);
-	}
-	
-	public Host(String hostname, String ipAddress, int port) {
-		this(hostname, ipAddress, port, Status.Down);
-	}
-	
-	public Host(String hostname, String ipAddress, Status status) {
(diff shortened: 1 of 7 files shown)


ipapapa:
The test that fails is unrelated and it succeeds on my workstation...

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
```diff
@@ -97,9 +98,22 @@ public String getRack() {
 	
 	public Host setRack(String rack) {
 		this.rack = rack;
+        setDatacenter(rack);
```
shailesh33: indentation

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
```diff
@@ -97,9 +98,22 @@ public String getRack() {
...
+	public String getDatacenter() {
+		return datacenter;
+	}
+	
+	/** 
+	 * Datacenter format us-east-x, us-west-x etc.
+	 * 
+	 */
+	private void setDatacenter(String rack) {
+		this.datacenter = rack.substring(0, rack.length() - 1);
```
shailesh33: please verify if that is the right approach

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HttpEndpointBasedTokenMapSupplier.java
```diff
@@ -40,109 +39,160 @@
...
-	private static final String DefaultServerUrl = "http://{hostname}:8080/REST/v1/admin/cluster_describe";
-	private final String serverUrl;
-	private static final Integer NumRetries = 2;
+    private static final Logger Logger = LoggerFactory.getLogger(HttpEndpointBasedTokenMapSupplier.class);
 
-	public HttpEndpointBasedTokenMapSupplier(int port) {
-		this(DefaultServerUrl, port);
+    private static final String DefaultServerUrl = "http://{hostname}:8080/REST/v1/admin/cluster_describe";
+    private final String serverUrl;
+    private static final Integer NumRetriesSameNode = 2;
```
shailesh33: First character should be small according to rest of the code

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HttpEndpointBasedTokenMapSupplier.java
```diff
@@ -40,109 +39,160 @@
...
-				in.close();
-			}
+    /**
+     * Finds a random host from the set of active hosts to perform
+     * cluster_describe
+     * 
+     * @param activeHosts
+     * @return a random host
+     */
+    private String getRandomHost(Set<Host> activeHosts) {
```
shailesh33: getRandomHostAddress.... we are returning address not Host

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
ipapapa: fixed

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
ipapapa: Great advice. Updated the `ConfigUtils` helper class to accommodate this case.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HttpEndpointBasedTokenMapSupplier.java
ipapapa: That was there before, but I agree so made the change.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/utils/ConfigUtils.java
```diff
@@ -16,37 +16,56 @@
...
 
-        return az;
+	return az;
     }
 
+    /**
+     * 
+     * @return the datacenter that the client is in
+     */
     public static String getDataCenter() {
```
ipapapa: @shailesh33 changed the `getDatacenter()` implementation

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
```diff
@@ -17,150 +17,153 @@
...
+	setDatacenter(rack);
+	return this;
+    }
+
+    public String getDatacenter() {
+	return datacenter;
+    }
+
+    private void setDatacenter(String rack) {
+	this.datacenter = ConfigUtils.getDataCenter(rack);
```
ipapapa: @shailesh33 used the `ConfigUtils` as advised.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HttpEndpointBasedTokenMapSupplier.java
ipapapa: This was there before, so I followed the conventions however, according the [Java Language Specifications](http://docs.oracle.com/javase/specs/#9367) _The names of constants in interface types should be, and final variables of class types may conventionally be, a sequence of one or more words, acronyms, or abbreviations, all uppercase, with components separated by underscore "_" characters_. SO updated.

timiblossom: Yeah, but you should group "static" variables in one place - don't mix them like that

ipapapa: @timiblossom I did not change that, it was like that before.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/utils/ConfigUtils.java
```diff
@@ -16,37 +16,56 @@
...
-        return dc;
+    }
+
+    /**
+     * 
+     * Datacenter format us-east-x, us-west-x etc.
+     * @param rack
+     * @return the datacenter based on the provided rack
+     */
+    public static String getDataCenter(String rack) {
```
shailesh33: [OPTIONAL] This should have been getDataCenterFromRack.

PR #133, opened 2016-09-14
Dual Writes - fixes broken model

jcacciatore:
Previously the prod connection pool was injected into the dual writer client. Now the connection pool to the shadow cluster is injected.


ipapapa:
LGTM

PR #139, opened 2016-09-28
maven-publish gradle bug workaround. 

anguenot:
Fix generated POMs to mark dependencies scope as gradle defined 'compile', not 'runtime'.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,6 +33,23 @@ subprojects {
         testCompile "junit:junit:4.11"
     }
 
+    // Maven-publish gradle bug workaround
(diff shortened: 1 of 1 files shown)


rspieldenner:
You shouldn't switch every library into the maven compile scope in the pom. Only libraries in the public api that people would then have to search for. The `compileApi` configuration added by a nebula plugin is a little less boilerplate.

Personal opinion is that the way maven leaks compile scope dependencies everywhere is a detriment.

ipapapa:
We can re-open this PR, depending on how #141 works for Maven users.

PR #141, opened 2016-09-28
Fix maven generated POM compile scopes.

anguenot:
Fix generated POMs to mark dependencies scope as gradle defined 'compile', not 'runtime'. Adding 'nebula' 'compileApi' scope.


rspieldenner:
Converting everything to `compileApi` is bogus. `compileApi` should be used for libraries that you have put into your api. The other libraries you need to compile are not needed for other users of your library to compile.

PR #152, opened 2016-10-28
Documentation cleanup for config settings

jcacciatore:


ipapapa:
LGTM, I am assuming 'getPoolShutdownDelay' is not used anywhere that is why we are removing it.

PR #153, opened 2016-10-30
Regression Bug Fix: Host object equality contract changed

jcacciatore:
The `equals()` and `hashcode()` contract changed to include all fields of the object which broke the token aware load balancer initialization. Unfortunately the breakage is very subtle; no errors are reported however the resulting topology is completely broken.

I did not revert the breaking change; this set of changes fixes the code that deserializes the json topology into proper Host objects w/all fields assigned. 

I updated the unit test to validate all fields when deserializing Host objects. The reason this wasn't caught in a unit test before is b/c although the test was updated it did not take into account all fields, only fields required by the previous equality contract.

### DIFF
--- a/dyno-core/src/test/java/com/netflix/dyno/connectionpool/impl/lb/AbstractTokenMapSupplierTest.java
+++ b/dyno-core/src/test/java/com/netflix/dyno/connectionpool/impl/lb/AbstractTokenMapSupplierTest.java
@@ -53,16 +53,16 @@ public class AbstractTokenMapSupplierTest {
 
 		List<Host> hostList = new ArrayList<>();
 
-		hostList.add(new Host("ec2-54-237-143-4.compute-1.amazonaws.com", 11211,"us-east-1d",  Status.Up));
-		hostList.add(new Host("ec2-50-17-65-2.compute-1.amazonaws.com", 11211, "us-east-1d" , Status.Up));
-		hostList.add(new Host("ec2-54-83-87-174.compute-1.amazonaws.com", 11211, "us-east-1c", Status.Up));
-		hostList.add(new Host("ec2-54-81-138-73.compute-1.amazonaws.com", 11211, "us-east-1c", Status.Up));
-		hostList.add(new Host("ec2-54-82-176-215.compute-1.amazonaws.com", 11211, "us-east-1c", Status.Up));
-		hostList.add(new Host("ec2-54-82-83-115.compute-1.amazonaws.com", 11211, "us-east-1e", Status.Up));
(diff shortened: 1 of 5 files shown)


dyno-core/src/test/java/com/netflix/dyno/connectionpool/impl/lb/AbstractTokenMapSupplierTest.java
```diff
@@ -51,15 +51,15 @@ public String getTopologyJsonPayload(String hostname) {
...
 
 		hostList.add(new Host("ec2-54-237-143-4.compute-1.amazonaws.com", 11211,"us-east-1d",  Status.Up));
 		hostList.add(new Host("ec2-50-17-65-2.compute-1.amazonaws.com", 11211, "us-east-1d" , Status.Up));
 		hostList.add(new Host("ec2-54-83-87-174.compute-1.amazonaws.com", 11211, "us-east-1c", Status.Up));
 		hostList.add(new Host("ec2-54-81-138-73.compute-1.amazonaws.com", 11211, "us-east-1c", Status.Up));
 		hostList.add(new Host("ec2-54-82-176-215.compute-1.amazonaws.com", 11211, "us-east-1c", Status.Up));
 		hostList.add(new Host("ec2-54-82-83-115.compute-1.amazonaws.com", 11211, "us-east-1e", Status.Up));
-		hostList.add(new Host("ec2-54-211-220-55.compute-1.amazonaws.com", 11211, "us-east-1c", Status.Up));
+		hostList.add(new Host("ec2-54-211-220-55.compute-1.amazonaws.com", 11211, "us-east-1e", Status.Up));
 		hostList.add(new Host("ec2-54-80-65-203.compute-1.amazonaws.com", 11211, "us-east-1c", Status.Up));
```
ipapapa: this should be us-east-1e.

alosix:
Looking over this.. I think this fix broke the functionality added in the PR that created the issue. I'm finding that with the latest builds it doesn't respect hosts with multiple dyno's running on different ports.

ipapapa:
The initial PR that effectively added the functionality of multiple Dyno ports heavily broke Dyno. We had spent significant time to debug and identify the issue. There was another PR https://github.com/Netflix/dyno/pull/165 that further addressed the issue of Dyno not being able to connect to a new node after a Dynomite node termination. If you wish to re-add the functionality please file a PR with proper test coverage.

alosix:
Indeed, so far I'm finding that with just a subclass of the host object putting back in the equals/compare/etc.. it seems to work well for my case still. That case seems far enough outside of the 'normal' usage of dyno that it might just be worth me adding that into the contrib package instead of the main section.

ipapapa:
I am not sure how it will work in the contrib package, but I can check it.

PR #157, opened 2016-12-16
Remove port from hashCode()

ipapapa:
The `hashCode()` has caused several issues when a custom port is provided by the user as the `DEFAULT_PORT=8102`. 

Looked into different options, like leveraging Archaius for "port override" but the problem is that we are creating the Host object twice. Once in the `HostSupplier` and once in the `TokenMapSupplier`. `EurekaHostSupplier` does not have the ability to return the port number.
Therefore, there is no reason to include in the comparison the port number.

### DIFF
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
@@ -22,7 +22,12 @@ import com.netflix.dyno.connectionpool.impl.utils.ConfigUtils;
 /**
  * Class encapsulating information about a host.
  *
- * This is immutable except for the host status
+ * This is immutable except for the host status.
+ * Note that the HostSupplier may not know the Dynomite port,
+ * whereas the Host object created by the load balancer
+ * may receive the port the cluster_describe REST call.
+ * Hence, we must not use the port in the equality and hashCode
+ * calculations.
  *
  * @author poberai
  * @author ipapapanagiotou
@@ -32,7 +37,7 @@ public class Host implements Comparable<Host> {
 
     public static final int DEFAULT_PORT = 8102;
     public static final Host NO_HOST = new Host("UNKNOWN", "UNKNOWN", 0, "UNKNOWN");
-    
+
     private final String hostname;
     private final String ipAddress;
     private final int port;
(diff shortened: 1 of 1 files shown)


ipapapa:
Test cases succeed locally, so the test failure is this common issue we have seen in the past.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
```diff
@@ -122,13 +122,18 @@ public InetSocketAddress getSocketAddress() {
...
+     * and therefore the hashCode() may fail if the user provides his
+     * own port. Hence, we do not have to include the port number in the
+     * hashCode() calculation.
+     */
     @Override
     public int hashCode() {
         final int prime = 31;
         int result = 1;
         result = prime * result + ((hostname == null) ? 0 : hostname.hashCode());
         result = prime * result + ((rack == null) ? 0 : rack.hashCode());
```
shailesh33: Do we have a use case of running multiple dynomite nodes on teh same node even if it means for testing purpose?

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
```diff
@@ -122,13 +122,18 @@ public InetSocketAddress getSocketAddress() {
         return socketAddress;
     }
 
+    /**
+     * The HostSupplier may not have the ability to return the port
+     * and therefore the hashCode() may fail if the user provides his
```
jcacciatore: The ```hashCode()``` doesn't 'fail', but it does cause equality checks to in fail in collections between Host objects created from the ```HostSupplier``` which may not know the dynomite port and the Host objects created by the load balancer. 

Consider rewording to state something to that effect, rather than hashCode failing.

jcacciatore: Also consider adding a note about this in the class javadoc, after the mention of a Host being immutable except for the status

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
ipapapa: We do not and even if we wanted the whole process would be fairly complicated to implement. Eureka provides a set of IP addresses and these are cross-validated with Florida's `cluster_describe` hostnames/tokens. Every hostname maps to a token, hence we have to change a lot of pieces to make a hostname mapping multiple tokens.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
```diff
@@ -22,7 +22,12 @@
 /**
  * Class encapsulating information about a host.
  *
- * This is immutable except for the host status
+ * This is immutable except for the host status.
+ * Note that the HostSupplier may not know the Dynomote port,
```
jcacciatore: Dynomite is misspelled (replace 'o' with 'i')

PR #160, opened 2017-01-02
Dynomite api support 

tiodollar:
Added support to dynomite REST API to search the cluester tokens https://github.com/Netflix/dynomite/wiki/REST.

### DIFF
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/AbstractTokenMapSupplier.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/AbstractTokenMapSupplier.java
@@ -123,6 +123,12 @@ public abstract class AbstractTokenMapSupplier implements TokenMapSupplier {
         localDatacenter = ConfigUtils.getDataCenter();
         unsuppliedPort = port;
     }
+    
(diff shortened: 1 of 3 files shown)


ipapapa:
@tiodollar this seems to fail the test cases. Any idea why? What is the case that this is needed?

tiodollar:
@ipapapa With the merge, All checks have passed.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/AbstractTokenMapSupplier.java
```diff
@@ -123,6 +123,12 @@ public AbstractTokenMapSupplier(int port) {
         localDatacenter = ConfigUtils.getDataCenter();
         unsuppliedPort = port;
     }
+    
+    public AbstractTokenMapSupplier(int port,String localZone,String localDatacenter) {
```
ipapapa: How do we need this extra constructor? I do not see it being used in this PR.

ipapapa:
It would be nice to add a few test cases to make sure that we have proper coverage as this is a new feature. I believe we have the HTTP based token map supplier tests that you can leverage. An example would be to create an exemplary JSON response from Dynomite and use that as the basis for testing.

PR #161, opened 2017-01-02
Added support to eval operation.

tiodollar:
Added support to eval operation, with that is possible to execute lua scripts.


ipapapa:
IMHO, we need some unit testing for this...

maksimlikharev:
is this planned for the new release, we are interested in this as well

ipapapa:
That is a good question that @tiodollar can answer. I do not have any problem adding this in Dyno as long as we have a proper testing coverage or at least some integration done through `dyno_test`. Dynomite has some specific limitations on `EVAL` and `EVALSHA` that are mentioned in the [notes](https://github.com/Netflix/dynomite/blob/dev/notes/redis.md).

tiodollar:
@ipapapa @maksimlikharev I'll work on this, give me some days!

ipapapa:
For an end to end Lua support we need:
* Both `EVAL` and `EVALSHA`. The tricky part is that there is no key so we need a way to map the EVAL to the correct node. Dyno uses token awareness after hashing the key with the proper node. I think having this PR do just that would be great.
* `SCRIPT LOAD` so that the client can programmatically load the LUA script. This is not currently supported by Dynomite, but adding it at the server side should be straightforward.

maksimlikharev:
any progress on that? @tiodollar
it seems we need it for real now, so I either submit my own PR or we sort that one rather quick

ipapapa:
EVAL support has been added to Dyno https://github.com/Netflix/dyno/pull/193 Since this PR has been fairly inactive and there are several conflicts, I am going to close it. If you still want to continue working it, then we can reopen and fix the conflicts.

PR #168, opened 2017-03-22
Implement mget

shailesh33:
This patch adds support for mget in the client.
It doesnt do the real "scatter gather" of mget yet.


ipapapa:
* Is there a specific reason on why compression was not added? What happens if somebody uses MSET with the compression FP enabled, and then tries to do a simple GET?
* Can you shrink the number of commits?

shailesh33:
@ipapapa the branch was merged /updated from master so I cannot merge the commits easily.

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
```diff
@@ -2430,8 +2515,31 @@ public Long pexpire(String key, long milliseconds) {
...
+        if (CompressionStrategy.NONE == connPool.getConfiguration().getCompressionStrategy()) {
+
+            return connPool.executeWithFailover(new MultiKeyOperation<List<String>>(new ArrayList<String>(Arrays.asList(keys)), OpName.MGET) {
+                @Override
+                public List<String> execute(Jedis client, ConnectionContext state) {
+                    return client.mget(keys);
+                }
+            });
+        } else {
+            return connPool.executeWithFailover(new CompressionValueMultiKeyOperation<List<String>>(new ArrayList<String>(Arrays.asList(keys)), OpName.MGET) {
```
jcacciatore: Same comment as above

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
```diff
@@ -2430,8 +2515,31 @@ public Long pexpire(String key, long milliseconds) {
...
 
     @Override
-    public List<String> mget(String... keys) {
-        throw new UnsupportedOperationException("not yet implemented");
+    public List<String> mget(String... keys) { return d_mget(keys).getResult(); }
+
+    public OperationResult<List<String>> d_mget(final String... keys) {
+        if (CompressionStrategy.NONE == connPool.getConfiguration().getCompressionStrategy()) {
+
+            return connPool.executeWithFailover(new MultiKeyOperation<List<String>>(new ArrayList<String>(Arrays.asList(keys)), OpName.MGET) {
```
shailesh33: took care of it.

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
shailesh33: done

shailesh33:
Thank you @jcacciatore for your review. I appreciate you taking the time.
The mget receives a null for a (nil) for a key that is missing. So that works.
If one of the node is down, an exception is thrown by the client.

jcacciatore:
Cool, looks good @shailesh33. The only other suggestion I have is to document the mget behavior, specifically for failure scenarios, in a javadoc header. Other than that LGTM

jcacciatore:
Didn't mean to close the PR...

PR #170, opened 2017-04-25
Bugfix: scan can go into an infinite loop when used with multi-az cluster

jcacciatore:
The bug fix isn't with the cursor handling directly, it's with other contributing factors such as the retry policy (if `RetryNTimes` is used), not taking any action when the local rack is not picked up, not being able to set the count as a scan parameter, etc.

This PR also modifies the demo test app. I split out the properties it expects into a file and some enforcement. Also added the ability to toggle between netflix prod + test for discovering dynomite clusters.

### DIFF
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/RetryNTimes.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/RetryNTimes.java
@@ -16,11 +16,12 @@
 package com.netflix.dyno.connectionpool.impl;
 
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 
 import com.netflix.dyno.connectionpool.RetryPolicy;
 
 /**
- * Simple implementation of {@link RetryPolicy} that ensures an operation can be re tried at most N times. 
+ * Simple implementation of {@link RetryPolicy} that ensures an operation can be re tried at most N times.
  * 
  * Note that RetryNTimes (2) means that a total of 2 + 1 = 3 attempts will be allowed before giving up.
  * 
@@ -30,7 +31,7 @@ import com.netflix.dyno.connectionpool.RetryPolicy;
 public class RetryNTimes implements RetryPolicy {
 
 	private int n; 
-	private final AtomicInteger count = new AtomicInteger(0);
+	private final AtomicReference<RetryState> state = new AtomicReference<>(new RetryState(0, false));
 	private final boolean allowCrossZoneFallback;
 	
 	public RetryNTimes(int n, boolean allowFallback) {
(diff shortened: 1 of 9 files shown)


dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
```diff
@@ -712,13 +712,13 @@ private String getCursorValue(final ConnectionContext state, final CursorBasedRe
         return "0";
     }
 
-    private List<OperationResult<ScanResult<String>>> scatterGatherScan(final CursorBasedResult<String> cursor, final String... pattern) {
+    private List<OperationResult<ScanResult<String>>> scatterGatherScan(final CursorBasedResult<String> cursor, final int count, final String... pattern) {
```
ipapapa: So we are adding `count` although it was not there before? Are we adding a new API call?

jcacciatore: Yes, but only for dyno-client. The `count` parameter is part of the redis api and is supported by jedis. Without this change the default of 10 is always used.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/RetryNTimes.java
```diff
@@ -44,22 +45,31 @@ public void begin() {
 
 	@Override
 	public void success() {
-		count.incrementAndGet();
+		boolean success = false;
```
ipapapa: what is the purpose of this change?

dyno-demo/src/main/java/com/netflix/dyno/demo/redis/DynoJedisDemo.java
```diff
@@ -223,10 +224,12 @@ public void runScanTest(boolean populateKeys) throws Exception {
 
         CursorBasedResult<String> cbi = null;
         do {
-            cbi = client.dyno_scan(cbi, "DynoClientTest_key-*");
-
-            logger.info("result: " + cbi.getResult().toString());
+			cbi = client.dyno_scan(cbi, 100, "DynoClientTest_key-*");
```
ipapapa: there are 500 keys written above, but you are doing count in 100. Is this correct?

dyno-demo/src/main/java/com/netflix/dyno/demo/redis/DynoJedisDemo.java
```diff
@@ -562,36 +565,39 @@ private String constructRandomValue(int sizeInKB) {
...
-		String url = "http://discovery.cloudqa.netflix.net:7001/discovery/v2/apps/" + clusterName;
+
+		String env = System.getProperty("netflix.environment", "test");
+		String discoveryKey = String.format("dyno.demo.discovery.%s", env);
+
+		if (!System.getProperties().containsKey(discoveryKey)) {
+			throw new IllegalArgumentException("Discovery URL not found");
+		}
+
+		final String url = String.format("http://%s/%s", System.getProperty(discoveryKey), clusterName);
```
ipapapa: Interesting, you are constructing the URL based on the properties... Looks good!

ipapapa:
LGTM!

PR #172, opened 2017-05-09
Use eureka client

twicksell:
Use EurekaClient interface instead of DiscoveryClient type (which implements EurekaClient)
Updated some outdated dependencies that would have come along with the new eureka-client

### DIFF
--- a/dyno-contrib/src/main/java/com/netflix/dyno/contrib/EstimatedHistogramBasedCounter.java
+++ b/dyno-contrib/src/main/java/com/netflix/dyno/contrib/EstimatedHistogramBasedCounter.java
@@ -83,6 +83,11 @@ public abstract class EstimatedHistogramBasedCounter extends AbstractMonitor<Num
 			return estHistogram.mean();
 		}
 
+        @Override
+        public Number getValue(int pollerIndex) {
+            return estHistogram.mean();
+        }
+
 		public void add(long n) {
             this.estHistogram.add(n);
         }
@@ -90,6 +95,7 @@ public abstract class EstimatedHistogramBasedCounter extends AbstractMonitor<Num
         public void reset() {
             this.estHistogram.getBuckets(true);
         }
+
 	}
 
 	public static class EstimatedHistogramPercentile extends EstimatedHistogramBasedCounter {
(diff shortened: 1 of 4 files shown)


dyno-contrib/src/main/java/com/netflix/dyno/contrib/EstimatedHistogramBasedCounter.java
```diff
@@ -83,13 +83,19 @@ public Number getValue() {
 			return estHistogram.mean();
 		}
 
+        @Override
+        public Number getValue(int pollerIndex) {
```
shailesh33: whats poller index supposed to be?

dyno-contrib/src/main/java/com/netflix/dyno/contrib/EurekaHostsSupplier.java
```diff
@@ -127,7 +134,7 @@ public String getApplicationName() {
         return applicationName;
     }
 
-    public DiscoveryClient getDiscoveryClient() {
+    public EurekaClient getDiscoveryClient() {
```
shailesh33: will this cause compilation errors or warnings?

dyno-contrib/src/main/java/com/netflix/dyno/contrib/EstimatedHistogramBasedCounter.java
twicksell: Seems to be a new concept added in Servo to keep track of the poller interval when recording more advanced metrics. I believe @brharrington could give a better explanation and help with a better implementation if anyone thinks there is value in it.

dyno-contrib/src/main/java/com/netflix/dyno/contrib/EurekaHostsSupplier.java
twicksell: If someone assigns the return value to the DiscoveryClient type, they would get a compilation error.

brharrington: Even if they assign to `EurekaClient` it could cause problems if someone just gets the new binary without recompiling. The return type is part of the signature used for looking up the method.

dyno-contrib/src/main/java/com/netflix/dyno/contrib/EstimatedHistogramBasedCounter.java
brharrington: It is something that should just be an implementation detail of servo. This could get refactored more, but most of the changes I would prefer would break compatibility of the metrics produced. I'm not sure how of a concern that is or how much use this data gets used internally.

This should be fine for now and we can explore it further if there is enough demand.

PR #177, opened 2017-06-26
Adding support for set with nx, ex and time

ipapapa:
Rob requested to support `String set(String key, String value, String nxxx, String expx, long time)`

### DIFF
--- a/dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
+++ b/dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
@@ -1319,11 +1319,6 @@ public class DynoJedisClient implements JedisCommands, BinaryJedisCommands, Mult
         return d_set(key, value).getResult();
     }
 
-    @Override
-    public String set(String key, String value, String nxxx, String expx, long time) {
-        throw new UnsupportedOperationException("not yet implemented");
-    }
-
     public OperationResult<String> d_set(final String key, final String value) {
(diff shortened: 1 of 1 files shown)


robzienert:
Thanks! Nice to get a view into what's necessary to add client support for maybe next time. :)

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
```diff
@@ -1344,6 +1339,31 @@ public String execute(final Jedis client, final ConnectionContext state) throws
...
+                @Override
+                public String execute(Jedis client, ConnectionContext state) throws DynoException {
+                    return client.set(key, value, nxxx, expx, time);
+                }
+            });
+        } else {
+            return connPool.executeWithFailover(new CompressionValueOperation<String>(key, OpName.SET) {
+                @Override
+                public String execute(final Jedis client, final ConnectionContext state) throws DynoException {
+                    return client.set(key, compressValue(value, state), nxxx, expx, time);
```
shailesh33: We need to check if dynomite has the appropriate parsing logic in place for this command.

ipapapa: It does 

````
redis-cli -p 8102 set mykey "hello" ex 100 NX
OK
````

shailesh33: Thanks.

PR #162, opened 2017-01-03
Added support for consul service discovery

tiodollar:
Added support to consul discovery, https://www.consul.io/

PR #178, opened 2017-06-27
Feature/consul support test

pcting:
Forked off of tiodollar:consul_support and added an integration test by instantiating an embedded Consul server. Netflix/dyno#162
Rebased against the latest on master (Netflix/dyno@9a56ca416ee6f6f449db99d1929edf681f98eaf2)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -62,6 +62,11 @@ project(':dyno-contrib') {
         compileApi "com.netflix.archaius:archaius-core:0.7.5"
         compileApi "com.netflix.servo:servo-core:0.12.16"
         compileApi 'com.netflix.eureka:eureka-client:1.6.2'
+        compileApi 'org.apache.commons:commons-lang3:3.5'
+        compileApi ('com.ecwid.consul:consul-api:1.2.1') {
+            exclude group: 'org.apache.httpcomponents'
+        }       
+        testCompile 'com.pszymczyk.consul:embedded-consul:0.3.3'
     }
 }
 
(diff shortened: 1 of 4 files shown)


pcting:
@ipapapa @tiodollar let me know what you guys think

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/AbstractTokenMapSupplier.java
```diff
@@ -123,6 +123,12 @@ public AbstractTokenMapSupplier(int port) {
         localDatacenter = ConfigUtils.getDataCenter();
         unsuppliedPort = port;
     }
+    
+    public AbstractTokenMapSupplier(int port,String localZone,String localDatacenter) {
```
ipapapa: where is this constructor used in this PR?

pcting: i accidentally pulled in additional commit. let me clean this up a bit

pcting:
@ipapapa i cleaned up the extra commits; let me know how it looks

tiodollar:
@ipapapa  @pcting Evething is ok to me, now! without this constructor!

ipapapa:
Approved and merged, but currently, DynoJedisClient only uses `EurekaHostSupplier` https://github.com/Netflix/dyno/blob/9a56ca416ee6f6f449db99d1929edf681f98eaf2/dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java#L3462. From that perspective this does not break anything but on the hand it requires programmatic changes to make it work. My recommendation would be to add somehow that one can choose which HostSupplier to provide.

PR #179, opened 2017-06-30
Update ConnectionPoolImpl.java

Aatish04:
Not able to connect to dynomite if first host is down because even if another host returns the results we are not resetting "lastException" variable, we end up throwing exception.


jcacciatore:
I recommend including a unit test for point fixes such as this.

Out of scope for this change, however, a more robust way to implement this would be to rely only on `RetryPolicy` instead of a special purpose variable. Unfortunately `RetryPolicy` does not support retrieval of the last failure (currently you can only mark that a failure occurred and get the attempts) but we should change this to facilitate getting rid of the `lastException` variable altogether.

PR #183, opened 2017-08-04
Support Redis auth

grfeng:


ipapapa:
Even if we add support for auth, Dynomite does not currently support it. Thoughts?

shailesh33:
auth requires dynomite support. Do you plan to add that too?

ipapapa:
Closing this due to inactivity and lack of description. Glad to reopen it once things improve.

mgddp:
Please can you re-open this pull request. I would like to make similar changes to support REDIS auth.  Should I create another pull request ?

ipapapa:
@mgddp I do not believe you can continue this PR so it might worth opening a new one with proper description. SSL support was added in https://github.com/Netflix/dyno/pull/184 Please check that one as well.

mgddp:
@ipapapa Thanks.  Yes, I am head of master and see the SSL support added.  We may just set a SSL Socket factory ( hostname veritifcation, certificate chain , trust etc ). However, I am not sure how REDIS auth "password" is being taken care in here ,if at all  ?

Also, do I need to create an issue for this  ? Or just fork and create a pull request.

PR #184, opened 2017-08-19
Feature/ssl support

KowalczykBartek:
### Motivations
Because of for many users authentication for database is necessary we should support some way to provide such security level for Dynomite - if not, lot of possible adopters would give up on adopting Dynomite. 

### Solution
Least invasive way to secure Dynomite is hiding behind NGINX and use NGINX's SSL termination in addition with required client-side-certificate. It should not impact big performance degradation.

### Setup
Setup of such secured ENV is described here https://gist.github.com/KowalczykBartek/37c882941e6827130350175bb07a7bcf

As well you can check **testSSLJedisClient** in **com.netflix.dyno.jedis.JedisConnectionFactoryIntegrationTest** to have overview how to configure client itself and even run simple Netty server that will allow you to tests SSL handshake **com.netflix.dyno.jedis.utilsMockedRedisResponse** (I had no idea for more eloquent name ^^ )

P.S
as you can see, I added  'maven-publish' plugin - when you type 
`./gradlew publishToMavenLocal `
all artifact will be uploaded to your local maven - no manual jars copying required anymore ;D


dyno-jedis/src/main/java/com/netflix/dyno/jedis/JedisConnectionFactory.java
```diff
@@ -42,47 +38,64 @@
...
 		private DynoConnectException lastDynoException;
-		
+
 		public JedisConnection(HostConnectionPool<Jedis> hostPool) {
 			this.hostPool = hostPool;
 			Host host = hostPool.getHost();
-			jedisClient = new Jedis(host.getHostAddress(), host.getPort(), hostPool.getConnectionTimeout(),
-					hostPool.getSocketTimeout());
+
+			if(sslSocketFactory == null)
```
ipapapa: Is setting `sslSocketFactory` configurable?

KowalczykBartek: hmm what does it mean ? generally if you pass me ```sslSocketFactory``` via builder I will further pass it to Jedis constructor - if not, then ```sslSocketFactory``` will be null, so just non-ssl-aware constructor will be used :)

ipapapa:
Can you add some description in the PR above? I know we chatted on Gitter but it may be worth adding the information for the rest of the reviewers as well as a link to your blog post for a broader bigger picture.

jcacciatore:
@KowalczykBartek has any performance testing been done ?

KowalczykBartek:
@jcacciatore performance testing of java' SSLEngine ? hmm, no - but as far as I know it is quite slow comparing e.g. to netty's openssl implementation - and can generate a lot of jitter (this is why Apple has its own) #lotofirony

If you mean  NGINX performance overhead - I didn't perform such tests as well - reason is quite simple - we had to do that if we wanted adopt Dynomite (security requirements) and there was no visible problems with latency.

KowalczykBartek:
any plans to merge ? can I help here somehow ?

dyno-jedis/src/test/java/com/netflix/dyno/jedis/utils/EmbeddedRedisInitializer.java
```diff
@@ -0,0 +1,43 @@
...
+        this.useSsl = useSsl;
+        this.response = response;
+    }
+
+    @Override
+    protected void initChannel(final SocketChannel ch) throws Exception {
+        final ChannelPipeline pipeline = ch.pipeline();
+
+        if (useSsl) {
+            final SSLEngine sslEngine = sslContext.createSSLEngine();
```
russell-lewis: While setting up the server & client not a part of this library or PR, it would nice to set the integration test server and client up to only use TLS 1.2.

dyno-jedis/src/test/java/com/netflix/dyno/jedis/utils/SSLContextUtil.java
```diff
@@ -0,0 +1,36 @@
...
+        final KeyStore trustStore = KeyStore.getInstance("jks");
+        trustStore.load(inputStream, STOREPASS.toCharArray());
+
+        final KeyManagerFactory keyManagerFactory = KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm());
+        keyManagerFactory.init(trustStore, STOREPASS.toCharArray());
+
+        final TrustManagerFactory trustManagerFactory = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+        trustManagerFactory.init(trustStore);
+
+        final SSLContext sslContext = SSLContext.getInstance("TLS");
```
russell-lewis: While setting up the server & client not a part of this library or PR, it would nice to set the integration test server and client up to only use TLS 1.2.

KowalczykBartek: can we treat JedisConnectionFactoryIntegrationTest.java as integration test ? there is full ssl connection established.

ipapapa: @KowalczykBartek my understanding is that @russell-lewis is more worried about the TLS version. Would this improve the TLS version `SSLContext sslContext = SSLContext.getInstance("TLSv1.2");`?

KowalczykBartek: Updated.

P.S 
even is "TLS" is used, server will respond with highest possible for client and server version.

dyno-jedis/src/test/java/com/netflix/dyno/jedis/utils/EmbeddedRedisInitializer.java
KowalczykBartek: Both client and server uses 
```
final SSLContext sslContext = SSLContext.getInstance("TLSv1.2");
```

But one tests failed - can you access this file from failed build  ?
```
file:///home/travis/build/Netflix/dyno/dyno-core/build/reports/tests/test/index.html
```
or at least run build again ? I am pretty sure that this failed randomly, locally it pass.

PR #186, opened 2017-09-12
Hashtag support for Dyno

ipapapa:
This is a live PR that I will be adding a few components. I would like some baseline review feedback. The current limitation of Dyno is that it does not allow to perform pipelines across different keys (i.e. a pipeline can be extended across operations of the same key). That reduces the performance of some key projects include [CloudDriver](https://github.com/spinnaker/clouddriver) and [dyno-queues](https://github.com/netflix/dyno-queues).

To get around this limitation we want to add the support of [Hash Tags](https://github.com/Netflix/dynomite/blob/dev/notes/recommendation.md#hash-tags) such that we can do a pipeline operation in the value defined in the hashtag. This will bypass the limitation of Dyno that pipelines can be done in the same key as pipelines will be created on the hashtag, and therefore significantly improve the performance.

* To ensure the proper support of hashtags we have to make sure they are enabled at the server side. The first component is for Dyno to get information of the hashtag delimiter from [Dynomite-manager](https://github.com/Netflix/dynomite-manager). Dynomite-manager will write in Dynomite's YML the supported hashtag. See the `HttpEndpointBasedTokenMapSupplier`.
* Currently, in a normal operation when a hashtag is defined at the Dynomite layer, we will try to extra the key from the hashtag. For example, if the key is `foo-{bar}`, we will use the `bar` as the value on which to perform the murmur hash function and determine the node to send the data.
* The guarantees are that a hashtag can only be two characters, i.e. `{}` or `[]`. Dynomite-manager provides these guarantees. From that we perform `StringUtils.substringBetween(key,Character.toString(hashtag.charAt(0)), Character.toString(hashtag.charAt(1)));` to extract hash value to use to get the token (test cases added).
* In a pipeline operation we have added the proper checks such that when there is a hashtag we make sure that check that each pipeline uses the same hashed value (not the same key).

A few other vertical improvements:
* updated Eureka/Servo dependencies and the Gradle wrapper


ipapapa:
@robzienert correct the client will provide whatever hashtag it needs to provide. The hashtag delimiter will be defined by Dynomite, so in the above case `{}` (missed to mention delimiter in my description and I fixed it).

dyno-core/src/main/java/com/netflix/dyno/connectionpool/BaseOperation.java
```diff
@@ -16,24 +16,36 @@
...
+
+    /**
+     * The key for the operation. Useful for implementing token aware routing.
+     * 
+     * @return String
+     */
+    public String getKey();
+
+    /**
+     * The hashtag for the operation. Useful for bypassing token aware routing.
```
shailesh33: this is not bypassing token aware routing

ipapapa: Good point, I will update the comment.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/TokenAwareSelection.java
```diff
@@ -32,77 +32,96 @@
...
+    public HostConnectionPool<CL> getPoolForOperation(BaseOperation<CL, ?> op) throws NoAvailableHostsException {
+
+        String key = op.getKey();
+        String hashtag = op.getHashtag();
+
+        HostToken hToken = null;
+        if (hashtag == null) {
+            hToken = this.getTokenForKey(key);
+        } else {
+            hToken = this.getTokenForKey(hashtag);
```
shailesh33: How will this work? From what I undestand hashtag is something like this "{}",
you should find out the content of the hashed key and use it.

dyno-core/src/test/java/com/netflix/dyno/connectionpool/impl/lb/TokenAwareSelectionHastagTest.java
```diff
@@ -0,0 +1,169 @@
...
+        return new BaseOperation<Integer, Long>() {
+
+            @Override
+            public String getName() {
+                return "TestOperation" + n;
+            }
+
+            @Override
+            public String getKey() {
+                return n + "-{bar}";
```
ipapapa: @robzienert here is the test that covers providing the hashtag in the key.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
```diff
@@ -44,39 +45,53 @@
...
+        this(hostname, ipAddress, port, rack, ConfigUtils.getDataCenterFromRack(rack), Status.Down, null);
     }
 
     public Host(String hostname, String ipAddress, String rack, Status status) {
-        this(hostname, ipAddress, DEFAULT_PORT, rack, ConfigUtils.getDataCenterFromRack(rack), status);
+        this(hostname, ipAddress, DEFAULT_PORT, rack, ConfigUtils.getDataCenterFromRack(rack), status, null);
     }
 
-    public Host(String name, String ipAddress, int port, String rack, String datacenter, Status status) {
+    public Host(String hostname, String ipAddress, String rack, Status status, String hashtag) {
```
ipapapa: @jcacciatore @shailesh33 here we create a host object with the `hashtag`.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallback.java
```diff
@@ -337,13 +340,27 @@ public void initWithHosts(Map<Host, HostConnectionPool<CL>> hPools) {
...
 		for (Host host : hPools.keySet()) {
 			String dc = host.getRack();
 			if (localRack != null && !localRack.isEmpty() && dc != null && !dc.isEmpty() && !localRack.equals(dc)) {
 				remoteDCs.add(dc);
 			}
+			/**
+			 * Checking for defined hashtags if all of them are the same.
+			 * If not we need to throw an exception.
+			 */
+			if(this.hashtag != null) {
```
ipapapa: We initialize hosts and we iterate over all the `Hosts` and make sure that all hashtags across all hosts are the same.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/TokenAwareSelection.java
```diff
@@ -32,77 +34,95 @@
...
+    @Override
+    public HostConnectionPool<CL> getPoolForOperation(BaseOperation<CL, ?> op, String hashtag) throws NoAvailableHostsException {
+
+        String key = op.getKey();
+
+        HostToken hToken = null;
+        if (hashtag == null) {
+            hToken = this.getTokenForKey(key);
+        } else {            
+            String hashValue = StringUtils.substringBetween(key,Character.toString(hashtag.charAt(0)), Character.toString(hashtag.charAt(1)));
```
ipapapa: @robzienert @shailesh33 @jcacciatore IF there is a hashtag defined, we use the substring inside the hashtag to determine the key to use for getting the node with the proper token.

shailesh33: as discussed verify substringBetween does the right thing for us

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallback.java
shailesh33: negative test case for different hashtags

shailesh33: check for empty hashtags

ipapapa:
Exposing the hashtag from Dynomite to Dyno through Dynomite-manager: https://github.com/Netflix/dynomite-manager/pull/79

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisPipeline.java
```diff
@@ -61,63 +62,120 @@
...
-                        public String getKey() {
-                            return key;
-                        }
-                    });
-                } catch (NoAvailableHostsException nahe) {
-                    cpMonitor.incOperationFailure(connection != null ? connection.getHost() : null, nahe);
-                    discardPipelineAndReleaseConnection();
-                    throw nahe;
+        // Get hashtag from the first host of the active pool
+        String hashtag = connPool.getActivePools().iterator().next().getHost().getHashtag();
```
ipapapa: @shailesh33 @jcacciatore To get the hashtag we select the first one from the active pool. We work under the assumption that all nodes will contain the same hashtag.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallback.java
```diff
@@ -207,7 +210,7 @@ private boolean attemptFallback() {
 			try {
 				
 				HostConnectionPool<CL> fallbackHostPool = 
-						(op != null) ? remoteDCSelector.getPoolForOperation(op) : remoteDCSelector.getPoolForToken(token);
+						(op != null) ? remoteDCSelector.getPoolForOperation(op,hashtag) : remoteDCSelector.getPoolForToken(token);
```
shailesh33: not related to this PR but is this really a "remote DC selector"?

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallback.java
```diff
@@ -337,13 +340,27 @@ public void initWithHosts(Map<Host, HostConnectionPool<CL>> hPools) {
...
+			 * Checking for defined hashtags if all of them are the same.
+			 * If not we need to throw an exception.
+			 */
+			if(this.hashtag != null) {
+			   String hashtagNew = host.getHashtag();
+			   if (!this.hashtag.equals(hashtagNew)){
+		               throw new RuntimeException("Hashtags are different across hosts");
+			   }
+			   this.hashtag = hashtagNew;
+			}
```
shailesh33: what happens when the first hashtag is null and subsequent are not? 
What you really want is a "set" data structure and make sure at the end the cardinality is 1.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/TokenAwareSelection.java
```diff
@@ -32,77 +34,96 @@
...
+            hToken = this.getTokenForKey(hashValue);
+        }
+
+        HostConnectionPool<CL> hostPool = null;
+        if (hToken != null) {
+            hostPool = tokenPools.get(hToken.getToken());
+        }
+        
+        if (hashtag == null || hashtag.isEmpty()) {
```
shailesh33: you don't really need this if. Throw a single exception with all the details

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallback.java
ipapapa: There is going to be a change in the code. But to explain this case, If the first hashtag is null, and the subsequent host hash a defined hashtag a `RuntimeException` will be thrown.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/TokenAwareSelection.java
ipapapa: fixed it in the upcoming commit.

ipapapa:
Made further tests with 100 keys and everything looks good. There is some issue with the Pipeline operations monitor that will be addressed in a separate PR. This is considered stable for review.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
```diff
@@ -152,6 +174,12 @@ public boolean equals(Object obj) {
             return false;
 
         Host other = (Host) obj;
+        /* we need a way to pass the information about the hashtag from the token map supplier
+         * to the host object. 
+         */
+        if (other.hashtag != null) {
```
shailesh33: why are we setting the hashtag here? Shouldnt equals be readonly?

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
```diff
@@ -152,6 +174,12 @@ public boolean equals(Object obj) {
...
+            setHashtag(other.hashtag);
+        }
         boolean equals = true;
```
shailesh33: Also shouldn't Equals and compareTo be logically be the same?

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
```diff
@@ -171,6 +199,10 @@ public int compareTo(Host o) {
 
     @Override
     public String toString() {
+        if (this.hashtag  != null){
```
shailesh33: Try combining the two returns. You do not really require a if else here.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/AbstractTokenMapSupplier.java
```diff
@@ -218,12 +227,13 @@ private boolean isLocalDatacenterHost(Host host) {
                 String zone = (String) jItem.get("zone");
                 String datacenter = (String) jItem.get("dc");
                 String portStr = (String) jItem.get("port");
+                String hashtag = (String) jItem.get("hashtag");
```
shailesh33: Just curious, what happens if there is no hashtag field from token map supplier? Have you tested it.?

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallback.java
```diff
@@ -328,28 +335,41 @@ public void initWithHosts(Map<Host, HostConnectionPool<CL>> hPools) {
...
+	        this.hashtag = allHostTokens.get(0).getHost().getHashtag();
+	        
+	        // Update inner state with the host tokens.
 		for (HostToken hToken : allHostTokens) {
+		          
+	                /**
+	                 * Checking for defined hashtags if all of them are the same.
+	                 * If not we need to throw an exception.
+	                 */
+		        String hashtagNew = hToken.getHost().getHashtag();
```
shailesh33: Think through all the cases when you write code and review it. Imagine a situation like this
3 hosts:
1) hashtag = null.
2) hashtag = {}
3) hashtag = {}

What will this do?

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/RoundRobinSelection.java
```diff
@@ -48,7 +48,7 @@ public RoundRobinSelection() {
 	}
 
 	@Override
-	public HostConnectionPool<CL> getPoolForOperation(BaseOperation<CL, ?> op) throws NoAvailableHostsException {
+	public HostConnectionPool<CL> getPoolForOperation(BaseOperation<CL, ?> op, String hashtag) throws NoAvailableHostsException {
```
shailesh33: What does the new hashtag parameter do here in this function? Why did you add it?

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/TokenAwareSelection.java
```diff
@@ -32,77 +37,88 @@
...
+    @Override
+    public HostConnectionPool<CL> getPoolForOperation(BaseOperation<CL, ?> op, String hashtag) throws NoAvailableHostsException {
+
+        String key = op.getKey();
+
+        HostToken hToken = null;
+        if (hashtag == null || hashtag.isEmpty()) {            
+            hToken = this.getTokenForKey(key);
+        } else {  
+            String hashValue = StringUtils.substringBetween(key,Character.toString(hashtag.charAt(0)), Character.toString(hashtag.charAt(1)));
```
shailesh33: To me, it seems that for Dynomite (https://github.com/Netflix/dynomite/blob/v0.5.9/src/dyn_message.c#L710) the key will include the hashtag. So for example, for a key ```foo{bar}foo``` and hashtag ```{}```, Dynomite will say the key is ```{bar}``` whereas your code will say the key is ```bar```.
Can you please check if I am correct?

ipapapa: If the hashtag is defined as `{}` why would Dynomite say the key is `{bar}`? This is incorrect.

shailesh33:
In that case, you will have to fix Dynomite side too.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
ipapapa: There are two pieces of information that we need to combine to determine healthy nodes.The host supplier (Eureka) and the host pool (provided by DM). The host supplier does not have information about the hashtags. Hence, in the simple operations, we use the host pool to derive the hashtags. Then we pass the hashtags to the `getPoolForOperation` such that we can identify the node based on the hashed value.

However, the Pipeline operations have been developed in a different way. A pipeline selects directly a Host object. The Host object is created based on the Host supplier, which does not have the hashtags (Eureka does not provide hashtags). I could not find a way to let the host pool inject it in the Host object since the equality is the only call happening to combine these two.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
ipapapa: `equals()` uses both the hostname and the rack. `compareTo()` either the hostname or the rack.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
ipapapa: Would this be better

```
    @Override
    public String toString() {
        
        String host ="Host with Hashtag [hostname=" + hostname + ", ipAddress=" + ipAddress + ", port=" + port + ", rack: " + rack
                + ", datacenter: " + datacenter + ", status: " + status.name();
        if (this.hashtag  != null){
            return host + ", hashtag=" + hashtag + "]";
        }
        return host + "]";
    }
```

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/AbstractTokenMapSupplier.java
ipapapa: I am not following the question. If no hashtag is provided, then it will work as before. Did I test it? Yes, I made sure all the vanilla test cases work. The idea is for everything to be backward compatible.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallback.java
ipapapa: Will fix this.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/RoundRobinSelection.java
ipapapa: Currently nothing. If an app uses `RoundRobinSelection` then hashtags are of no meaning. I could probably through an exception if the Dynomite cluster has a hashtag defined and the app uses `RoundRobinSelection`. What do you think?

shailesh33: no, why would you throw an error for RoundRobin with hashtags?

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallback.java
ipapapa: I think this is a wrong name. Looking in the code `String dc = host.getRack();`... and then `HostSelectionStrategy<CL> remoteSelector = remoteDCSelectors.get(dc);`. I can fix this.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/RoundRobinSelection.java
ipapapa: Because in RoundRobin you are not using the keys to determine where the data are going to be sent, so hashtags as part of the key have no meaning. In RoundRobin, Dyno uses a `circularList`. Now let us assume that sb configures Dynomite to support hashtags, then Dyno and Dynomite would be misaligned.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
shailesh33: I am only saying making changes while calling "compareTo" or Equals is not good.
Those functions should not have any sideeffects

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
shailesh33: String host ="Host [hostname=" + hostname + ", ipAddress=" + ipAddress + ", port=" + port + ", rack: " + rack
                + ", datacenter: " + datacenter + ", status: " + status.name() + ", hashtag=" + hashtag + "]";

Test it!

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/RoundRobinSelection.java
shailesh33: not sure what you mean here. but that's ok.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallback.java
ipapapa: Refactored the names. They will appear in the upcoming commit.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
ipapapa: Done!

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
ipapapa: You are right, let me think about it. I have committed all the rest of the fixes.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/TokenAwareSelection.java
shailesh33: correct. My bad!

dyno-core/src/main/java/com/netflix/dyno/connectionpool/Host.java
ipapapa: Fixed it

shailesh33:
looks good to me.

PR #187, opened 2017-09-18
Support for REDIS Auth

mgddp:
Allow REDIS Auth to be configured as part of host configuration. Configure JEDIS client to set Auth if available in host configuration.


ipapapa:
That code would probably never be executed because nobody is setting the password. Any thoughts about that?

mgddp:
Plan to use it with netflix conductor.

ipapapa:
My comment was about the code not being executed, not how you plan to use it. This PR also needs test case coverage.

davidwadden:
This PR can be closed as #244 has been merged.

PR #190, opened 2017-11-02
Compression demo test

shailesh33:
Add a basic compression test in the demo app with hmset/hmget in a pipeline.


dyno-demo/src/main/java/com/netflix/dyno/demo/redis/DynoJedisDemo.java
```diff
@@ -922,6 +954,10 @@ public static void main(String args[]) throws IOException {
                 demo.runSScanTest(true);
                 break;
             }
+            case 9: {
+                    demo.runCompressionTest();
```
ipapapa: `runCompressionInPipeline`?

ipapapa:
Do we also need a unit test for compression and pipelines or is this enough?

shailesh33:
This is not really a unit test. It does not run as part of the build. This is a demo app that we can run locally as part of dev test and is subject to change. I am committing the code just for reference if we need it next time.

shailesh33:
I have renamed the test function

PR #192, opened 2017-11-16
empty token map supplier

ipapapa:


dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
```diff
@@ -3630,10 +3630,16 @@ private void setHashtagConnectionPool(HostSupplier hostSupplier, ConnectionPoolC
             List<HostToken> hostTokens;
             if (tokenMapSupplier != null) {
                 hostTokens = tokenMapSupplier.getTokens(hostSet);
+                /* Dyno cannot reach the TokenMapSupplier endpoint,
```
shailesh33: should this exception be throw in tokenMapSupplier.getTokens?

ipapapa: So you are asking to move it in the [AbstractTokenMapSupplier](https://github.com/Netflix/dyno/blob/master/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/AbstractTokenMapSupplier.java#L140-L159). If the TokenMapSupplier endpoint cannot be reached then we have a [warning](https://github.com/Netflix/dyno/blob/master/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/AbstractTokenMapSupplier.java#L155). If we move this no hosts in the TokenMapSupplier `error` to be thrown in the AbstractTokenMapSupplier, then we have to error out the above warning.

PR #193, opened 2017-11-17
Support for EVAL command for Dynomite

shailesh33:


dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
```diff
@@ -429,6 +431,62 @@ public Long execute(Jedis client, ConnectionContext state) {
         });
     }
 
+    @Override
+    public Object eval(String script, int keyCount, String... params) { return d_eval(script, keyCount, params).getResult(); }
+
+    public OperationResult<Object> d_eval(final String script, final int keyCount, final String... params) {
+        if (keyCount == 0) {
+            throw new DynoException("Need atleast one key in script");
```
ipapapa: `at least`

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
```diff
@@ -429,6 +431,62 @@ public Long execute(Jedis client, ConnectionContext state) {
...
+        throw new UnsupportedOperationException("This function is Not Implemented. Please use eval instead.");
+    }
+
+    @Override
+    public Object evalsha(String script) {
+        throw new UnsupportedOperationException("This function is Not Implemented. Please use eval instead.");
+    }
+
+    @Override
+    public Boolean scriptExists(String sha1) {
```
ipapapa: What is the issue and we cannot implement `scriptExists`?

shailesh33: which node will you ask if a script exists? there is no key in a script till it is executed

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
shailesh33: Done

PR #199, opened 2017-12-15
Binary support

ipapapa:
* Adding support for binary commands in Dyno.
* Binary commands under pipelines are also supported.

Limitations: The binary support does not include compression and does not support hashtags. The latter requires a string key. Not all APIs have been added. We can add them gradually based on customer usage.

### DIFF
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/hash/Murmur1HashPartitioner.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/hash/Murmur1HashPartitioner.java
@@ -40,6 +40,14 @@ public class Murmur1HashPartitioner implements HashPartitioner {
         byte[] b = bb.array();
         return UnsignedIntsUtils.toLong(Murmur1Hash.hash(b, b.length));
 	}
+    
+	@Override
+	public Long hash(byte[] key) {
+		if (key == null) {
+			return 0L;
+		}
+        return UnsignedIntsUtils.toLong(Murmur1Hash.hash(key, key.length));
+	}
 	
     @Override
 	public Long hash(long key) {
@@ -63,4 +71,6 @@ public class Murmur1HashPartitioner implements HashPartitioner {
 	public HostToken getToken(Long keyHash) {
 		throw new RuntimeException("NotImplemented");
 	}
+
+
 }
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/hash/Murmur2HashPartitioner.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/hash/Murmur2HashPartitioner.java
@@ -62,6 +62,14 @@ public class Murmur2HashPartitioner implements HashPartitioner {
 		byte[] b = bb.array();
 		return  UnsignedIntsUtils.toLong(Murmur2Hash.hash32(b, b.length));
 	}
+	
+	@Override
+	public Long hash(byte[] key) {
+		if (key == null) {
+			return 0L;
+		}
+		return  UnsignedIntsUtils.toLong(Murmur2Hash.hash32(key, key.length));
+	}
 
 	@Override
 	public HostToken getToken(Long keyHash) {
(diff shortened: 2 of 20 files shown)


dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/hash/Murmur1HashPartitioner.java
```diff
@@ -40,6 +40,14 @@ public Long hash(String key) {
         byte[] b = bb.array();
         return UnsignedIntsUtils.toLong(Murmur1Hash.hash(b, b.length));
```
shailesh33: call the byte array implementation here

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/hash/Murmur2HashPartitioner.java
```diff
@@ -62,6 +62,14 @@ public Long hash(String key) {
 		byte[] b = bb.array();
 		return  UnsignedIntsUtils.toLong(Murmur2Hash.hash32(b, b.length));
```
shailesh33: same comment as above

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/hash/Murmur3HashPartitioner.java
```diff
@@ -62,6 +62,14 @@ public Long hash(String key) {
 		byte[] b = bb.array();
 		return UnsignedIntsUtils.toLong(Murmur3Hash.hash32(b, b.length));
```
shailesh33: same here

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/hash/Murmur1HashPartitioner.java
ipapapa: the byte[] implementation does not have the length.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/TokenAwareSelection.java
```diff
@@ -79,30 +77,46 @@ public Long get(HostToken x) {
...
 
     /**
-     * If a hashtag is provided by Dynomite then we use that to create the key to hash.
+     * Identifying the proper pool for the operation. A couple of things that may affect the decision
+     * (a) hashtags: In this case we will construct the key by decomposing from the hashtag
+     * (b) type of key: string keys vs binary keys. 
+     * In binary keys hashtags do not really matter.
      */
     @Override
     public HostConnectionPool<CL> getPoolForOperation(BaseOperation<CL, ?> op, String hashtag) throws NoAvailableHostsException {
```
shailesh33: I am no expert, but I am wondering if its a good idea to split the two implementations into something like:
StringBaseOperation &
BinaryBaseOperation? so there is not if else anywhere?

Compression is implemented inside them may be.

dyno-core/src/test/java/com/netflix/dyno/connectionpool/impl/ConnectionPoolImplTest.java
```diff
@@ -679,9 +679,14 @@ public String getName() {
 			
 
 			@Override
-			public String getKey() {
+			public String getStringKey() {
```
shailesh33: can there not be two functions getKey() in Java, one returning string and one returning Byte[]

ipapapa: I will complain about `Duplicate Method`.

dyno-core/src/test/java/com/netflix/dyno/connectionpool/impl/lb/TokenAwareSelectionBinaryTest.java
```diff
@@ -0,0 +1,268 @@
+/**
```
shailesh33: is this a new test file you added?

ipapapa: Yes

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/TokenAwareSelection.java
shailesh33: are you planning to do this? or should I merge?

ipapapa: I looked into it and it becomes very intrusive and it would require more code. We will need `StringBaseOperation`, `BinaryBaseOperation` and each one of them for the multikeys, then go to each command and use the corresponding one. Currently, the overhead of having a null byte array or a null string is very minimal. I suggest if we go in this direction to do it in a separate PR. Let us make sure that after releasing an RC, clients see a benefit.

PR #200, opened 2018-01-08
 BugFix in ConsulHostSupplier

srikanthm-1:
The if condition was comparing different enum instances.

### DIFF
--- a/dyno-contrib/src/main/java/com/netflix/dyno/contrib/consul/ConsulHostsSupplier.java
+++ b/dyno-contrib/src/main/java/com/netflix/dyno/contrib/consul/ConsulHostsSupplier.java
@@ -14,18 +14,11 @@
  * limitations under the License.
  */
 package com.netflix.dyno.contrib.consul;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-
-import org.apache.commons.lang3.StringUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import com.ecwid.consul.v1.ConsulClient;
 import com.ecwid.consul.v1.QueryParams;
 import com.ecwid.consul.v1.Response;
-import com.ecwid.consul.v1.agent.model.Check;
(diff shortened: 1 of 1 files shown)


srikanthm-1:
Hi,
If you are okay with this PR and when you merge it, can you also release it to maven?

Thanks

srikanthm-1:
Hi,
Can you please take a look at this when you get a chance? Its time sensitive since we are running in Production with this bug.

Thanks

dyno-contrib/src/main/java/com/netflix/dyno/contrib/consul/ConsulHostsSupplier.java
```diff
@@ -107,7 +107,7 @@ public Host apply(HealthService info) {
 
                         Host.Status status = Host.Status.Up;
                         for (com.ecwid.consul.v1.health.model.Check check : info.getChecks()) {
-                            if (check.getStatus().equals(Check.CheckStatus.CRITICAL)) {
+                            if (check.getStatus() == Check.CheckStatus.CRITICAL) {
```
ipapapa: In general, `==` checks if both objects point to the same memory location whereas .equals() evaluates to the comparison of values in the objects. Are you sure this is a correct fix?

srikanthm-1: Yes. The equals method in Enum.java does an `==` comparison underneath. http://hg.openjdk.java.net/jdk7/jdk7/jdk/file/ff9ad99b63cc/src/share/classes/java/lang/Enum.java#l134

In general to check for equality in enum, using `==` is a better option than the equals method since 
1) It will check the type of the enum compared are the same during compile time (Which would have prevented this bug) 
2) It will eliminate the possibility of NPE.

srikanthm-1: @ipapapa Can you also release it to maven?

PR #201, opened 2018-02-01
Read the Atomic reference variable in the loop to avoid cpu run off.

shailesh33:


dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/RetryNTimes.java
```diff
@@ -46,17 +46,17 @@ public void begin() {
 	@Override
 	public void success() {
 		boolean success = false;
-		RetryState rs = state.get();
 		while (!success) {
+			RetryState rs = state.get();
```
ipapapa:
```
RetryState rs = null;
while(!success) {
    rs = state.get();
....
```

shailesh33: why? rs is not being used outside the loop so does it matter?

ipapapa: because you re-creating the `RetryState` object in every iteration. There is some performance penalty for that. In the above approach, the object is being re-used in every iteration (not recreated).

PR #202, opened 2018-02-05
Support for configurable Dynomite port

shailesh33:
Remove the Default port from host object and put it in connectionPoolConfiguration. The idea is to make the port configurable.
All constructors of Host class take a port as parameter making it mandatory for the user to provide one.
It is the responsibility of the HostSupplier and TokenMapSupplier to provide hosts with the correct port. Hence the Default implementations in the code now take port as a parameter.
I also attempted to fix the SSL test with the correct datacenter provided so that it doesn't fail in calculating replication factor.

The builder object was not taking the tokenMapSupplier so user would create a connectionPoolConfiguration and provide tokenMapSupplier via it, many times, not using the ArchaiusConnectionPoolConfiguration.

### DIFF
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/ConnectionPoolConfigurationImpl.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/ConnectionPoolConfigurationImpl.java
@@ -31,7 +31,8 @@ import com.netflix.dyno.connectionpool.impl.utils.ConfigUtils;
 
 public class ConnectionPoolConfigurationImpl implements ConnectionPoolConfiguration {
 	
-	// DEFAULTS 
+	// DEFAULTS
+	public static final int DEFAULT_DYNOMITE_PORT = 8102;
 	private static final int DEFAULT_MAX_CONNS_PER_HOST = 3;
 	private static final int DEFAULT_MAX_TIMEOUT_WHEN_EXHAUSTED = 800;
 	private static final int DEFAULT_MAX_FAILOVER_COUNT = 3; 
@@ -56,6 +57,7 @@ public class ConnectionPoolConfigurationImpl implements ConnectionPoolConfigurat
   private HashPartitioner hashPartitioner;
 	private String hashtag;
 	private final String name;
+	private int port = DEFAULT_DYNOMITE_PORT;
 	private int maxConnsPerHost = DEFAULT_MAX_CONNS_PER_HOST; 
 	private int maxTimeoutWhenExhausted = DEFAULT_MAX_TIMEOUT_WHEN_EXHAUSTED; 
 	private int maxFailoverCount = DEFAULT_MAX_FAILOVER_COUNT; 
@@ -113,6 +115,7 @@ public class ConnectionPoolConfigurationImpl implements ConnectionPoolConfigurat
         this.localDataCenter = config.getLocalDataCenter();
         this.localRack = config.getLocalRack();
         this.localZoneAffinity = config.localZoneAffinity;
+		this.port = config.getPort();
         this.maxConnsPerHost = config.getMaxConnsPerHost();
         this.maxFailoverCount = config.getMaxFailoverCount();
         this.maxTimeoutWhenExhausted = config.getMaxTimeoutWhenExhausted();
@@ -133,6 +136,10 @@ public class ConnectionPoolConfigurationImpl implements ConnectionPoolConfigurat
 		return name;
 	}
 
+	@Override
+	public int getPort() {
+		return port;
+	}
 
 	@Override
 	public int getMaxConnsPerHost() {
@@ -247,6 +254,7 @@ public class ConnectionPoolConfigurationImpl implements ConnectionPoolConfigurat
 				", tokenSupplier=" + tokenSupplier +
 				", hostConnectionPoolFactory=" + hostConnectionPoolFactory +
 				", name='" + name + '\'' +
+				", port=" + port +
 				", maxConnsPerHost=" + maxConnsPerHost +
 				", maxTimeoutWhenExhausted=" + maxTimeoutWhenExhausted +
 				", maxFailoverCount=" + maxFailoverCount +
@@ -274,6 +282,11 @@ public class ConnectionPoolConfigurationImpl implements ConnectionPoolConfigurat
 	}
 
 	// ALL SETTERS
+	public ConnectionPoolConfigurationImpl setPort(int port) {
+		this.port = port;
+		return this;
+	}
+
 	public ConnectionPoolConfigurationImpl setMaxConnsPerHost(int maxConnsPerHost) {
 		this.maxConnsPerHost = maxConnsPerHost;
 		return this;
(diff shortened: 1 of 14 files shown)


dyno-core/src/test/java/com/netflix/dyno/connectionpool/impl/lb/AbstractTokenMapSupplierTest.java
```diff
@@ -26,16 +26,16 @@
...
+			"\"{\"token\":\"3450843231\",\"hostname\":\"ec2-54-81-138-73.compute-1.amazonaws.com\",\"port\":\"8102\",\"dc\":\"us-east-1\",\"ip\":\"54.81.138.73\",\"zone\":\"us-east-1c\"},\""+
+			"\"{\"token\":\"587531700\",\"hostname\":\"ec2-54-82-176-215.compute-1.amazonaws.com\",\"port\":\"8102\",\"dc\":\"us-east-1\",\"ip\":\"54.82.176.215\",\"zone\":\"us-east-1c\"},\"" +
+			"\"{\"token\":\"3101134286\",\"hostname\":\"ec2-54-82-83-115.compute-1.amazonaws.com\",\"port\":\"8102\",\"dc\":\"us-east-1\",\"ip\":\"54.82.83.115\",\"zone\":\"us-east-1e\"},\"" +
+			"\"{\"token\":\"237822755\",\"hostname\":\"ec2-54-211-220-55.compute-1.amazonaws.com\",\"port\":\"8102\",\"dc\":\"us-east-1\",\"ip\":\"54.211.220.55\",\"zone\":\"us-east-1e\"},\"" +
+			"\"{\"token\":\"1669478519\",\"hostname\":\"ec2-54-80-65-203.compute-1.amazonaws.com\",\"port\":\"8102\",\"dc\":\"us-east-1\",\"ip\":\"54.80.65.203\",\"zone\":\"us-east-1e\"}]\"";
 
-	private TokenMapSupplier testTokenMapSupplier = new AbstractTokenMapSupplier() {
+	private TokenMapSupplier testTokenMapSupplier = new AbstractTokenMapSupplier(8102) {
```
ipapapa: Can we add a test to a different port? This will cover the non-default use case.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/ConnectionPoolConfigurationImpl.java
```diff
@@ -113,6 +115,7 @@ public ConnectionPoolConfigurationImpl(ConnectionPoolConfigurationImpl config) {
         this.localDataCenter = config.getLocalDataCenter();
         this.localRack = config.getLocalRack();
         this.localZoneAffinity = config.localZoneAffinity;
+		this.port = config.getPort();
```
ipapapa: indentation

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HttpEndpointBasedTokenMapSupplier.java
```diff
@@ -44,26 +44,26 @@
...
-    public HttpEndpointBasedTokenMapSupplier(int port) {
-	this(DefaultServerUrl, port);
+    public HttpEndpointBasedTokenMapSupplier(int httpPort, int dynomitePort) {
+	this(DefaultServerUrl, httpPort, dynomitePort);
     }
 
-    public HttpEndpointBasedTokenMapSupplier(String url, int port) {
-	super(port);
+    public HttpEndpointBasedTokenMapSupplier(String url, int httpPort, int dynomitePort) {
+	super(dynomitePort);
```
ipapapa: Are you sure that super should not be `super(httpPort)`? Was it a bug before?

dyno-core/src/test/java/com/netflix/dyno/connectionpool/impl/lb/AbstractTokenMapSupplierTest.java
```diff
@@ -26,16 +26,18 @@
...
-			"\"{\"token\":\"237822755\",\"hostname\":\"ec2-54-211-220-55.compute-1.amazonaws.com\",\"port\":\"11211\",\"dc\":\"us-east-1\",\"ip\":\"54.211.220.55\",\"zone\":\"us-east-1e\"},\"" +
-			"\"{\"token\":\"1669478519\",\"hostname\":\"ec2-54-80-65-203.compute-1.amazonaws.com\",\"port\":\"11211\",\"dc\":\"us-east-1\",\"ip\":\"54.80.65.203\",\"zone\":\"us-east-1e\"}]\"";
+	// For some negative test cases, let the json payload have some nodes with wrong port
+	final String json = "[{\"token\":\"3051939411\",\"hostname\":\"ec2-54-237-143-4.compute-1.amazonaws.com\",\"port\":\"22123\",\"dc\":\"us-east-1\",\"ip\":\"54.237.143.4\",\"zone\":\"us-east-1d\"}\"," +
+			"\"{\"token\":\"188627880\",\"hostname\":\"ec2-50-17-65-2.compute-1.amazonaws.com\",\"port\":\"22122\",\"dc\":\"us-east-1\",\"ip\":\"50.17.65.2\",\"zone\":\"us-east-1d\"},\"" +
+			"\"{\"token\":\"2019187467\",\"hostname\":\"ec2-54-83-87-174.compute-1.amazonaws.com\",\"port\":\"22122\",\"dc\":\"us-east-1\",\"ip\":\"54.83.87.174\",\"zone\":\"us-east-1c\" },\"" +
+			// TEST WRONG RACK here.
```
ipapapa: Should the comment be outside of the JSON document?

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
```diff
@@ -3938,6 +3961,7 @@ private DynoJedisClient buildDynoJedisClient() {
 
             try {
                 Logger.info("Starting connection pool for app " + appName);
+                Logger.info("Starting connection pool for app " + appName);
```
ipapapa: Is this a repetition of the above line?

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HttpEndpointBasedTokenMapSupplier.java
shailesh33: the port is not being used and it was passed unnecessary because http port should have no role in AbstractTokenMapSupplier. Http port is really a property of HttpEndPointBasedTokenMapSupplier.

dyno-core/src/test/java/com/netflix/dyno/connectionpool/impl/lb/AbstractTokenMapSupplierTest.java
shailesh33: its a code comment. Doesnt affect the json document. I purposely kept it there to highlight the suttle difference

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
shailesh33: Good eye.

dyno-core/src/test/java/com/netflix/dyno/connectionpool/impl/lb/AbstractTokenMapSupplierTest.java
ipapapa: I understand but I did not get why it has to be inside the JSON document. Your call...

shailesh33: Done

PR #205, opened 2018-02-14
Use Hosts from Token Map Supplier

shailesh33:
Previously we were taking the hosts from the HostSupplier and then using them to prime connections.
The hosts from the tokenMapSupplier were only used later to initialize the selection strategy.
However, the token map supplier is the real source of truth for all the information about hosts
like its port, secure port (coming up feature), peer port, token etc.

This change does the right thing by using hosts from HostSupplier only to determine the up and down hosts.
And uses the hosts from the TokenMapSupplier to prime connections since the Hosts from the
TokenMapSupplier has the port information.

Ideally we want to differentiate the Host object from the HostSupplier and TokenMapSupplier by
creating different classes for these two purposes.

### DIFF
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/HostSupplier.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/HostSupplier.java
@@ -15,7 +15,7 @@
  ******************************************************************************/
 package com.netflix.dyno.connectionpool;
 
-import java.util.Collection;
+import java.util.List;
 
 /**
  * Interface for a supplier of host objects that map to the dynomite cluster. The {@link ConnectionPool} object can use this to 
@@ -30,5 +30,5 @@ public interface HostSupplier {
 	 * Return a collection of dynomite hosts for the connection pool
 	 * @return Collection<Host>
 	 */
-	public Collection<Host> getHosts();
+	public List<Host> getHosts();
 }
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/HostSupplier.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/HostSupplier.java
@@ -27,8 +27,8 @@ import java.util.List;
 public interface HostSupplier {
 	
 	/**
-	 * Return a collection of dynomite hosts for the connection pool
-	 * @return Collection<Host>
+	 * Return a list of dynomite hosts for the connection pool
+	 * @return List<Host>
 	 */
 	public List<Host> getHosts();
 }
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostsUpdater.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostsUpdater.java
@@ -30,7 +30,7 @@ import org.slf4j.LoggerFactory;
 
 public class HostsUpdater {
 
-	private static final org.slf4j.Logger Logger = LoggerFactory.getLogger(ConnectionPoolImpl.class);
+	private static final Logger Logger = LoggerFactory.getLogger(ConnectionPoolImpl.class);
 
 	private final HostSupplier hostSupplier;
 	private final TokenMapSupplier tokenMapSupplier;
@@ -73,8 +73,8 @@ public class HostsUpdater {
 		}
 
 		/**
-		 * HostTracker should return the hosts that we get from TMS.
-		 * Hence get the hosts from HostSupplier and map them to TMS
+		 * HostTracker should return the hosts that we get from TokenMapSupplier.
+		 * Hence get the hosts from HostSupplier and map them to TokenMapSupplier
 		 * and return them.
 		 */
 		Collections.sort(allHosts);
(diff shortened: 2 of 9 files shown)


dyno-core/src/main/java/com/netflix/dyno/connectionpool/HostSupplier.java
```diff
@@ -30,5 +30,5 @@
 	 * Return a collection of dynomite hosts for the connection pool
 	 * @return Collection<Host>
```
ipapapa: Please update the Javadoc to contain `List` since this the structure of your preference...  Note that a `Collection` is more generic.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostsUpdater.java
```diff
@@ -15,54 +15,102 @@
  */
 package com.netflix.dyno.connectionpool.impl;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.List;
+import java.util.*;
```
ipapapa: Do we really need all the APIs from the util package?

dyno-core/src/main/java/com/netflix/dyno/connectionpool/HostSupplier.java
shailesh33: ok. The thing is we are everywhere taking a collection and converting it to a string etc.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostsUpdater.java
```diff
@@ -15,54 +15,102 @@
...
+import com.netflix.dyno.connectionpool.exception.DynoException;
 import com.netflix.dyno.connectionpool.exception.NoAvailableHostsException;
+import com.netflix.dyno.connectionpool.impl.lb.HostToken;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class HostsUpdater {
 
-	private final HostSupplier hostSupplier; 
+	private static final org.slf4j.Logger Logger = LoggerFactory.getLogger(ConnectionPoolImpl.class);
```
ipapapa: `org.slf4j.Logger` --> `Logger`. You have already imported the package above.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostsUpdater.java
shailesh33: editor does it, It is not really going to import all of them unless we use them

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostsUpdater.java
```diff
@@ -15,54 +15,102 @@
...
+		 * HostTracker should return the hosts that we get from TMS.
+		 * Hence get the hosts from HostSupplier and map them to TMS
+		 * and return them.
+		 */
+		Collections.sort(allHosts);
+		Set<Host> hostSet = new HashSet<>(allHosts);
+		// Create a list of host/Tokens
+		List<HostToken> hostTokens;
+		if (tokenMapSupplier != null) {
+			Logger.info("Getting Hosts from TMS");
```
ipapapa: Please write `TokenMapSupplier`...

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostsUpdater.java
```diff
@@ -15,54 +15,102 @@
...
+			hostTokens = tokenMapSupplier.getTokens(hostSet);
+
+			if (hostTokens.isEmpty()) {
+				throw new DynoException("No hosts in the TokenMapSupplier");
+			}
+		} else {
+			throw new DynoException("TokenMapSupplier not provided");
+		}
+
+		Map<Host, Host> allHostSetFromTMS = new HashMap<>();
```
ipapapa: We need to choose. If we are going to call it `TMS` or `TokenMapSupplier` in the code. Having both names might be difficult when searching in the code for names.

ipapapa: `<Host,Host>`?

dyno-core/src/main/java/com/netflix/dyno/connectionpool/HostSupplier.java
shailesh33: done

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostsUpdater.java
shailesh33: done

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostsUpdater.java
shailesh33: done

PR #206, opened 2018-02-14
MSET and MSETNX

ipapapa:
Adding support for multi-key operations like MSET and MSETNX.


dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
```diff
@@ -134,7 +134,7 @@ private MultiKeyOperation(final List<String> keys, final OpName o) {
             this.op = o;
         }
 
-        @Override
+		@Override
```
shailesh33: indentation. I just do not get what is the standard here.

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
```diff
@@ -2705,15 +2708,53 @@ public String get(String s) {
...
+            return connPool.executeWithFailover(new MultiKeyOperation<Long>(Arrays.asList(keysvalues), OpName.MSETNX) {
+                @Override
+                public Long execute(Jedis client, ConnectionContext state) {
+                    return client.msetnx(keysvalues);
+                }
+            });
+        } else {
+            return connPool.executeWithFailover(new CompressionValueMultiKeyOperation<Long>(Arrays.asList(keysvalues), OpName.MSETNX) {
+                   @Override
+                   public Long execute(final Jedis client, final ConnectionContext state) throws DynoException {
```
shailesh33: Where are we compressing the values here?

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
```diff
@@ -2705,15 +2708,53 @@ public String get(String s) {
...
+                @Override
+                public String execute(Jedis client, ConnectionContext state) {
+                    return client.mset(keysvalues);
+                }
+            });
+        } else {
+            return connPool.executeWithFailover(new CompressionValueMultiKeyOperation<String>(Arrays.asList(keysvalues), OpName.MSET) {
+                   @Override
+                   public String execute(final Jedis client, final ConnectionContext state) throws DynoException {
+                            return client.mset(keysvalues);
```
shailesh33: same, no compression happening here

dyno-core/src/main/java/com/netflix/dyno/connectionpool/MultiKeyCompressionOperation.java
```diff
@@ -0,0 +1,18 @@
...
+ * @author ipapapa
+ *
+ * @param <CL>
+ * @param <R>
+ */
+public interface MultiKeyCompressionOperation<CL, R> extends Operation<CL, R> {
+
+	 String[] compressMultiKeyValue(ConnectionContext ctx, String... value);
+
+    String decompressValue(ConnectionContext ctx, String value);
```
shailesh33: any reason why this doesnt take a list of strings?

ipapapa: In multi-key write operations, we receive `String... keyValues` which is a combination of keys and values and we return either a `String` or ` Long`. In multi-key read operations, the API receives a `String keys` which is only the keys that we want to be read and we return a List of values. The list of values is being transformed through the `CollectionUtils.transform` to each independent components (this is how you had set it up and it looks good to me). Then we take each value and we decompress. Hence, our `compressMultiKeyValue` has numerous inputs and has some logic to perform the compression like selecting every second element which is the values. The `decompressValue` does not have any special logic other than decompressing each value one-by-one

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
```diff
@@ -241,48 +240,56 @@ public String decompressValue(String value, ConnectionContext ctx) {
...
+                          if ((2 * value.length()) > connPool.getConfiguration().getValueCompressionThreshold()) {
+                              newItems.add(i, ZipUtils.compressStringToBase64String(value));
+                              ctx.setMetadata("compression", true);
+                          }
+                          
+                      } catch (IOException e) {
+                          Logger.warn(
+                                  "UNABLE to compress [" + value + "] for key [" + getStringKey() + "]; sending value uncompressed");
+                      }
+                	 }
```
shailesh33: indentation

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
```diff
@@ -241,48 +240,56 @@ public String decompressValue(String value, ConnectionContext ctx) {
...
-
-            return result;
+        public String[] compressMultiKeyValue(ConnectionContext ctx, String... keyValues) {
+        	 List<String> items = Arrays.asList(keyValues);
+           	 List<String> newItems = new ArrayList<String>();
+           	 
+              for (int i = 0 ; i < items.size() ; i++) {
+             	 if( i%2 == 0 ) {
+             		 String value = items.get(i);
+
```
shailesh33: please maintain that comment about 2

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
```diff
@@ -241,48 +240,56 @@ public String decompressValue(String value, ConnectionContext ctx) {
...
      * <lh>String Operations</lh>
      * <li>{@link #mget(String...) MGET}</li>
+     * <li>{@link #mset(String...) MSET}</li>
+     * <li>{@link #msetnx(String...) MSETNX}</li>
      * </ul>
      *
      * @param <T>
      *            the parameterized type
      */
     private abstract class CompressionValueMultiKeyOperation<T> extends MultiKeyOperation<T>
```
shailesh33: may be rename this to alternate value Multi key operation etc? just an idea

shailesh33:
LGTM

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
ipapapa: Fixed.

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
ipapapa: Done

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
ipapapa: It is not an alternate value. It is a key-value pair.

PR #208, opened 2018-02-15
ensure that gzip streams are closed

brharrington:
The GZIP streams in java use zlib internally and will
leak native memory if not closed.

/cc @dmuino

### DIFF
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/utils/ZipUtils.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/utils/ZipUtils.java
@@ -30,9 +30,9 @@ public final class ZipUtils {
 
     public static byte[] compressString(String value) throws IOException {
         ByteArrayOutputStream baos = new ByteArrayOutputStream(value.length());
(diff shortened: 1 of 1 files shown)


dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/utils/ZipUtils.java
```diff
@@ -30,9 +30,9 @@ private ZipUtils() {
 
     public static byte[] compressString(String value) throws IOException {
         ByteArrayOutputStream baos = new ByteArrayOutputStream(value.length());
-        GZIPOutputStream gos = new GZIPOutputStream(baos);
-        gos.write(Base64.encode(value.getBytes(StandardCharsets.UTF_8)));
-        gos.close();
+        try (GZIPOutputStream gos = new GZIPOutputStream(baos)) {
+            gos.write(Base64.encode(value.getBytes(StandardCharsets.UTF_8)));
+        }
```
ipapapa: where are you closing the `gos.close()`?

ipapapa: The `try` will close it...

PR #211, opened 2018-02-22
adding EXISTS and DEL w/ multiple keys

ipapapa:


shailesh33:
LGTM, but I think exists needs to be supported on the server side.

Did you test the rest?

PR #212, opened 2018-02-22
use Secure port for SSL given by TokenMapSupplier

shailesh33:
By default, secure port is the same as the dynomite listening port. This way, current OSS users using SSL will not break.
If the tokenMapSupplier provides a secure_port in the cluster_describe, that one will be used with SSL.

### DIFF
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/AbstractTokenMapSupplier.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/AbstractTokenMapSupplier.java
@@ -227,13 +227,19 @@ public abstract class AbstractTokenMapSupplier implements TokenMapSupplier {
                 String zone = (String) jItem.get("zone");
                 String datacenter = (String) jItem.get("dc");
                 String portStr = (String) jItem.get("port");
+                String securePortStr = (String) jItem.get("secure_port");
                 String hashtag = (String) jItem.get("hashtag");
                 int port = Host.DEFAULT_PORT;
                 if (portStr != null) {
                     port = Integer.valueOf(portStr);
                 }
 
-                Host host = new Host(hostname, ipAddress, port, zone, datacenter, Status.Up, hashtag);
+                int securePort = port;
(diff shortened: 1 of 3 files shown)


dyno-jedis/src/main/java/com/netflix/dyno/jedis/JedisConnectionFactory.java
```diff
@@ -81,7 +81,7 @@ public JedisConnection(HostConnectionPool<Jedis> hostPool) {
 						hostPool.getSocketTimeout());
 			}
 			else {
-				jedisClient = new Jedis(host.getHostAddress(), host.getPort(), hostPool.getConnectionTimeout(),
+				jedisClient = new Jedis(host.getHostAddress(), host.getSecurePort(), hostPool.getConnectionTimeout(),
```
ipapapa: Do we support secure port handling on Dynomite? Will this be half-baked if the server does not support it yet?

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/AbstractTokenMapSupplier.java
```diff
@@ -227,13 +227,19 @@ private boolean isLocalDatacenterHost(Host host) {
                 String zone = (String) jItem.get("zone");
                 String datacenter = (String) jItem.get("dc");
                 String portStr = (String) jItem.get("port");
+                String securePortStr = (String) jItem.get("secure_port");
```
ipapapa: Does DM currently expose the secure port?

shailesh33: You reviewed it :)

dyno-jedis/src/main/java/com/netflix/dyno/jedis/JedisConnectionFactory.java
shailesh33: what do you mean by half-baked? It depends on how cluster_describe says.

PR #215, opened 2018-03-14
Add support for zadd with parameters.

v1r3n:


dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisPipeline.java
```diff
@@ -2142,12 +2142,21 @@ public String get(String key, String val) {
```
shailesh33: Changes look good to me. I am wondering if this function is straightforward to be added too? otherwise go ahead and merge it.

PR #216, opened 2018-04-24
Added information to exceptions

shailesh33:


dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallback.java
```diff
@@ -123,7 +123,7 @@ public HostSelectionWithFallback(ConnectionPoolConfiguration config, ConnectionP
 	}
 
 	private Connection<CL> getConnection(BaseOperation<CL, ?> op, Long token, int duration, TimeUnit unit, RetryPolicy retry)
-            throws NoAvailableHostsException, PoolExhaustedException, PoolTimeoutException {
+            throws NoAvailableHostsException, PoolExhaustedException, PoolTimeoutException, PoolOfflineException {
```
ipapapa: What was the reason for adding the `PoolOfflineException`?

shailesh33: because we are throwing that exception on line 163 below? Is it not that the convention in Java ? correct me if I am wrong.

PR #220, opened 2018-05-17
Improve Scan operation reliability

shailesh33:
Dyno does a scatter gather of scan operation on the shards in a given rack. However this failed if one of the node in the AZ went down and scan never recovered. Also If a node in the current AZ was down, scan would never finish since it would keep on hoping between the other two racks all the time. This fix does the following:
- A tokenRackMapper interface is used to keep track of which token in pinned to which rack.
- The cursor that gets sent to the application (CursorBasedResultImpl) implements this interface
- The scatter-gather remembers which token is pinned to which rack in the cursor
- next time a scan is called (pagination), this TokenRackMapper is checked to see which rack the requests should be sent to.

This fix achieves the following behavior:
1) When a scan is first called, the scatter-gather will forward the requests to all the nodes in the local rack.  If the local rack is not defined, it will use a random rack as a local rack.
2) If a node is down in that rack, it will failover to a different replica in another rack for *that token only*. It remembers this in the TokenRackMapper. All subsequent calls to scan (pagination) with this cursor will use the same nodes that the previous operation used. That way, even if the node that was originally down comes back up, it will not be used.
3) While the scan iteration is going on, if one of the node goes now, the scan will throw a PoolOfflineOperation. 

This might not be good since if one is scanning the entire DB the results need to be thrown out. This is an inherent problem since the iterator returned by redis is very specific to a node and is not portable to another replica.

### DIFF
--- a/dyno-demo/src/main/java/com/netflix/dyno/demo/redis/DynoJedisDemo.java
+++ b/dyno-demo/src/main/java/com/netflix/dyno/demo/redis/DynoJedisDemo.java
@@ -108,7 +108,8 @@ public class DynoJedisDemo {
 
 			@Override
 			public List<Host> getHosts() {
-				return hosts;
+				return getHostsFromDiscovery(clusterName);
+//				return hosts;
 			}
 		};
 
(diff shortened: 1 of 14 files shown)


dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
```diff
@@ -793,7 +793,11 @@ private String getCursorValue(final ConnectionContext state, final CursorBasedRe
 
     private List<OperationResult<ScanResult<String>>> scatterGatherScan(final CursorBasedResult<String> cursor,
             final int count, final String... pattern) {
-        return new ArrayList<>(connPool.executeWithRing(new BaseKeyOperation<ScanResult<String>>("SCAN", OpName.SCAN) {
+
+        if (!(cursor instanceof TokenRackMapper)) {
```
ipapapa: When will this throw an exception? Is there a case that `cursor` is not an instance of `TokenRacMapper?

dyno-demo/src/main/java/com/netflix/dyno/demo/redis/DynoJedisDemo.java
```diff
@@ -983,7 +1001,7 @@ public static void main(String args[]) throws IOException {
 			}
 			case 5: {
 				final boolean writeKeys = Boolean.valueOf(props.getProperty("dyno.demo.scan.populateKeys"));
```
ipapapa: Apparently, `writeKeys` is read by the property file but not used as `demo.runScanTest(true)`, always defaults to `true`.

dyno-demo/src/main/java/com/netflix/dyno/demo/redis/DynoJedisDemo.java
```diff
@@ -292,16 +294,32 @@ public void runScanTest(boolean populateKeys) throws Exception {
...
 			List<String> results = cbi.getStringResult();
+			count += results.size();
+			int i = 0;
 			for (String res : results) {
-				logger.info(res);
+				logger.info("{}) {}", i, res);
+				i++;
 			}
-		} while (!cbi.isComplete());
+			Thread.sleep(1000);
```
ipapapa: You are trying to calculate how fast SCAN performs but at the same time you seem to be sleeping the thread in the loop.

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
shailesh33: Ideally the cursor is something that we create and send it out. so we do not expect user the create it. Apparently the API isnt good

dyno-demo/src/main/java/com/netflix/dyno/demo/redis/DynoJedisDemo.java
shailesh33: oh sorry about that.

dyno-demo/src/main/java/com/netflix/dyno/demo/redis/DynoJedisDemo.java
shailesh33: man. I was doing some testing and I left this. This is not really doing performance testing but gives more time for say taking a node down, etc

Issue #185, opened 2017-09-01
switch to jdk 8 as compile platform breaks 1.7 compatibility


maksimlikharev:
Altho you have
 
sourceCompatibility = 1.7
targetCompatibility = 1.7

but it seems artifacts now compiled with java 8, that produce  this particular error when code run with java 7.

```
java.lang.NoSuchMethodError: java.util.concurrent.ConcurrentHashMap.keySet()Ljava/util/concurrent/ConcurrentHashMap$KeySetView;
	at com.netflix.dyno.connectionpool.impl.hash.BinarySearchTokenMapper.initBinarySearch(BinarySearchTokenMapper.java:114)
	at com.netflix.dyno.connectionpool.impl.hash.BinarySearchTokenMapper.initSearchMecahnism(BinarySearchTokenMapper.java:78)
	at com.netflix.dyno.connectionpool.impl.lb.TokenAwareSelection.initWithHosts(TokenAwareSelection.java:66)
	at com.netflix.dyno.connectionpool.impl.lb.HostSelectionWithFallback.initWithHosts(HostSelectionWithFallback.java:348)
	at com.netflix.dyno.connectionpool.impl.ConnectionPoolImpl.initSelectionStrategy(ConnectionPoolImpl.java:627)
	at com.netflix.dyno.connectionpool.impl.ConnectionPoolImpl.start(ConnectionPoolImpl.java:526)
	at com.netflix.dyno.jedis.DynoJedisClient$Builder.startConnectionPool(DynoJedisClient.java:3521)
	at com.netflix.dyno.jedis.DynoJedisClient$Builder.createConnectionPool(DynoJedisClient.java:3509)
	at com.netflix.dyno.jedis.DynoJedisClient$Builder.buildDynoJedisClient(DynoJedisClient.java:3487)
	at com.netflix.dyno.jedis.DynoJedisClient$Builder.build(DynoJedisClient.java:3421)

``` 

if you continue to support 1.7, you have to compile with 1.7 jdk.

ipapapa:
That is a good point. We have to fix that.

ipapapa:
It has been addressed in #186.

PR #221, opened 2018-05-25
Mget binary support

balajivenki:
@ipapapa  and @shailesh33  I have implemented the mget with support for byte[] keys. Here are the summary of changes

- Implementing MultiKeyBinaryCommands interface and providing implementation for mget with byte[]
- Modified MultiKeyOperation to take binaryKey and assign it based on the first key in the List passed.
- The earlier assumption for String keys will not be null and now added null check
- CompressionValueMultiKeyOperation is modified to support changes from MultiKeyOperation
- I did a unit test with String mget and byte[] mget and the keys are span across multiple nodes. The result was correct.
- Now that we are implementing MultiKeyBinaryCommands this opens the door for byte[] implementation for other missing REDIS commands too and I will add them at later stage.

Please review the PR and let me know if anything needs to be added or modified.

### DIFF
--- a/dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
+++ b/dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
@@ -129,7 +129,10 @@ public class DynoJedisClient implements JedisCommands, BinaryJedisCommands, Mult
         private final OpName op;
 
         private MultiKeyOperation(final List keys, final OpName o) {
-            Object firstKey = (keys != null) ? keys.get(0) : null;
+            //filter nulls
+            keys.removeIf(Objects::isNull);
+
+            Object firstKey = (keys != null && keys.size() > 0) ? keys.get(0) : null;
 
             if(firstKey != null) {
                 if (firstKey instanceof String) {//string key
--- a/dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
+++ b/dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
@@ -47,7 +47,7 @@ import org.apache.commons.lang3.ArrayUtils;
 import static com.netflix.dyno.connectionpool.ConnectionPoolConfiguration.CompressionStrategy;
 
 public class DynoJedisClient implements JedisCommands, BinaryJedisCommands, MultiKeyCommands,
-        ScriptingCommands {
+        ScriptingCommands, MultiKeyBinaryCommands {
 
     private static final Logger Logger = org.slf4j.LoggerFactory.getLogger(DynoJedisClient.class);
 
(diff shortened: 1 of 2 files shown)


shailesh33:
@ipapapa please review this PR

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
```diff
@@ -127,9 +128,25 @@ public String getStringKey() {
...
+                    this.binaryKeys = null;
+                } else if (firstKey instanceof byte[]) {//binary key
+                    this.keys = null;
+                    this.binaryKeys = keys;
+                } else {//something went wrong here
+                    this.keys = null;
+                    this.binaryKeys = null;
+                }
+            } else {
+                this.keys = null;
```
ipapapa: Is there a case that the `firstKey` is `null` and the rest are not?

balajivenki: @ipapapa  good point. I have addd the null filter to address it and when all nulls sent it fails. Even the old implementation had this issue.

ipapapa:
Seems we are seeing the same issue in Travis as in https://github.com/Netflix/dyno/issues/222

balajivenki:
@ipapapa I think it is to do with nebula dependency. In my local I can see it can be fixed by upgrading nebula.netflixoss to 5.0.0

ipapapa:
Can you update Nebula in this PR to fix it?

balajivenki:
@ipapapa  Yes upgrading to nebula 5.0.0 fixed the issue. I have removed the single line I added for filtering null as it is causing an exception. I will add it though, rest changes stays the same. Between is dyno jdk 8.0 supported? I see https://github.com/Netflix/dyno/issues/185 is merged though. Please let me know. This is for using stream.filter.

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
balajivenki: @ipapapa  Even when we filter nulls the change for this is across all places where ever we use MultiKeyOperation and CompressionValueMultiKeyOperation and it will fail saying 

`redis.clients.jedis.exceptions.JedisDataException: value sent to redis cannot be null`

so its a known  behavior even with current implementation. If you want I can go ahead and add a Collections filter on all places before passing the keys down further. Please let me know.

balajivenki:
Thanks a lot @ipapapa  and @shailesh33

ipapapa:
Would you like me to draft an RC?

balajivenki:
Yes @ipapapa I was going to ask RC too.

ipapapa:
Since https://github.com/Netflix/dyno/pull/223 was fired at the same date. Let us wait to get this approved as well and I will draft an RC tomorrow.

balajivenki:
Thanks for the update @ipapapa

ipapapa:
Done: https://github.com/Netflix/dyno/releases/tag/v1.6.4-rc.3

balajivenki:
@ipapapa  and @shailesh33 Can you please push the new rc into maven repo?

We only have till rc.2 there(https://mvnrepository.com/artifact/com.netflix.dyno/dyno-core)

ipapapa:
It is there: https://bintray.com/netflixoss/oss-candidate/dyno/1.6.4-rc.3

balajivenki:
Not sure why I am not able to see it in maven repo website though.

image

PR #228, opened 2018-07-10
Renamed EC2_REGION and EC2_AVAILABILITY_ZONE

marimcmurtrie:
I renamed below 2 system properties names to be more appropriate.
EC2_REGION -> LOCAL_DATACENTER
EC2_AVAILABILITY_ZONE -> LOCAL_RACK

I ran tests in dyno-jedis. All tests pass.

### DIFF
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/utils/ConfigUtils.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/utils/ConfigUtils.java
@@ -25,13 +25,14 @@ package com.netflix.dyno.connectionpool.impl.utils;
 public class ConfigUtils {
 
     public static String getLocalZone() {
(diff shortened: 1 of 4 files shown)


shailesh33:
@marimcmurtrie this is a breaking change. Ideally you should maintain backward compatibility for this. One way is to check for EC2 variables and if not found then use new LOCAL* variables.

marimcmurtrie:
Oh right I thought about it but forgot to implement, thank you @shailesh33 ! I will add if- condition for backward compatibility.

marimcmurtrie:
My push failed the test but it looks like not related to my code change.
The failed test,  RateTrackerTest.java, passes locally. 

The error message is "[ERR] dns: error starting tcp server: accept tcp 127.0.0.1:44148: use of closed network connection" so is it some sort of network glitch?

PR #230, opened 2018-08-08
Remove usage of deprecated Guava methods.

jhspaybar:


dyno-contrib/src/main/java/com/netflix/dyno/contrib/EstimatedHistogramBasedCounter.java
```diff
@@ -15,6 +15,7 @@
  */
 package com.netflix.dyno.contrib;
 
+import com.google.common.base.MoreObjects;
 import com.google.common.base.Objects;
```
ipapapa: Since we are not using Objects anymore, shall we remove the `import` as well?

jhspaybar: It’s still in use for hashcode. It’s just the toString stuff that was moved.

Issue #234, opened 2018-10-05
Make circularList threadsafe and fix Integer overflow


kishorekasi:
add()/remove() methods are not write thread-safe. currentIndex in InnerList can overflow if not updated.

ipapapa:
@kishorekasi is this considered completed based on #235?

kishorekasi:
Yes, it’s complete. I will close the issue. Thanks @ipapapa

PR #235, opened 2018-10-05
Make circularList threadsafe (#234)

kishorekasi:
Turned add()/remove() to synchronized methods and modified the read index in CircularList a AtomicLong to avoid integer overflow.

### DIFF
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/CircularList.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/CircularList.java
@@ -18,7 +18,7 @@ package com.netflix.dyno.connectionpool.impl.lb;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 
 /**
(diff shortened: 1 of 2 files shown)


kishorekasi:
@ipapapa @smukil

kishorekasi:
@jcacciatore

smukil:
Late to the party, but I feel like we can avoid having to use the 'synchronize' keyword to have better throughput by using optimistic concurrency control.

Eg: addElement() can be rewritten like:

	public void addElement(T element) {
                 List<T> origList = ref.get().list;		
                 do {
                     origList = ref.get().list;
		     boolean isPresent = origList.contains(element);
         	     if (isPresent) {
			return;
		     }
                      List<T> newList = new ArrayList<T>(origList);
		      newList.add(element);
                 } while (!ref.compareAndSet(origList, newList));
	}

The other methods can be written similarly.

(Forgive the syntax errors if any)
This will allow for better parallelism especially in the cases where we're trying to add something that's already present, or remove something that's not already there.

@kishorekasi @ipapapa What do you think? Of course, some basic empirical analysis needs to be done to find out if it's actually better or not.

kishorekasi:
@smukil I think your solution is better compared to synchronizing add()/remove().

ipapapa:
I agree it might be better, the question we need to answer is whether the number of calls is that many that providing a better solution worth the effort. The Circular list is not called on per call basis but rather when we add/remove a connection. However, if the solution is really simple, I vote yes to it :)

kishorekasi:
Based on usage it may not warrant a change for now. We can implement @smukil’s optimization in future when needed. I will close the issue for now.

PR #239, opened 2018-10-12
Extend DynoClientDemo to test DynoDualWriterClient

kishorekasi:
Added support to test DynoDualWriterClient when dual writer support is enabled through demo.properties (dyno.demo.dualwrite.enabled=true). With dual writers DynoClientDemo will accept 3 command line parametes <primaryClusterName> <ShadowClusterName> <testNumber>.


ipapapa:
You have not yet filed a PR for the dual writer, therefore I am wondering why doesn't this go in the same PR with the dual write?

PR #240, opened 2018-10-15
Make CircularList::getNextElement() thread safe

smukil:
We missed making this method synchronous in the previous patch.


smukil:
Good point. Abandoning this change.

PR #241, opened 2018-10-15
Add debug logging to display hosts that went up/down

smukil:
We occasionally notice that some hosts are completely unreachable from
Dyno. This patch adds logging to display all the new hosts that came
up and previously active hosts that went down. This patch is meant to
help with debugging the next time the above issue shows up.

### DIFF
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostStatusTracker.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostStatusTracker.java
@@ -128,9 +128,20 @@ public class HostStatusTracker {
 	 * @return true/false indicating whether the set of hosts has changed or not.
 	 */
 	public boolean checkIfChanged(Collection<Host> hostsUp, Collection<Host> hostsDown) {
-		return activeSetChanged(hostsUp) || inactiveSetChanged(hostsUp, hostsDown);
+		boolean changed = activeSetChanged(hostsUp) || inactiveSetChanged(hostsUp, hostsDown);
+
+		if (changed && logger.isDebugEnabled()) {
+			Set<Host> changedHostsUp = new HashSet<>(hostsUp);
+			changedHostsUp.removeAll(activeHosts);
+			changedHostsUp.forEach(x -> logger.debug("New host up: {}", x.getHostAddress()));
+
+			Set<Host> changedHostsDown = new HashSet<>(hostsDown);
+			changedHostsDown.removeAll(inactiveHosts);
+			changedHostsDown.forEach(x -> logger.debug("New host down: {}", x.getHostAddress()));
+		}
+		return changed;
 	}
-	
+
 	/**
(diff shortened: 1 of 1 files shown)


dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostStatusTracker.java
```diff
@@ -128,9 +128,45 @@ public boolean inactiveSetChanged(Collection<Host> hostsUp, Collection<Host> hos
...
+	 * logs the difference in hosts.
+	 * @param hostsUp
+	 * @param hostsDown
+	 */
+	public void logChangedHosts(Collection<Host> hostsUp, Collection<Host> hostsDown) {
+		if (!logger.isDebugEnabled()) return;
+
+		Set<Host> changedHostsUp = new HashSet<>();
+		changedHostsUp.addAll(hostsUp);
+		for (Host host : activeHosts) {
```
kishorekasi: Will Java Set operations like https://docs.oracle.com/javase/7/docs/api/java/util/Set.html#removeAll makes this more terse? Few other operations that may help you here are retainAll/containsAll.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostStatusTracker.java
```diff
@@ -128,9 +128,45 @@ public boolean inactiveSetChanged(Collection<Host> hostsUp, Collection<Host> hos
...
+		Set<Host> changedHostsDown = new HashSet<>();
+		changedHostsDown.addAll(hostsDown);
+		for (Host host : inactiveHosts) {
+			if (changedHostsDown.contains(host)) {
+				changedHostsDown.remove(host);
+			}
+		}
+
+		for (Host host : changedHostsUp) {
+			logger.debug("New host up: " + host.getHostAddress());
```
kishorekasi: log4j best practices:

- Log4j logging can be efficient (lazy argument evaluation) when parameters can be printed with {} instead of concatenating the string representation of object. The above can be changed as - 

`if (logger.isDebugEnabled()) {
  logger.debug("New host up: {}", host.getHostAddress());
}
`

- It is also good practice to use isDebugEnabled() for debug logs as it can avoid expensive computation of log messages in the log4j stack.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostStatusTracker.java
smukil: Thanks! I've incorporated this to make it shorter.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostStatusTracker.java
smukil: Done, thanks!

Issue #237, opened 2018-10-12
Support pipeline operations from JedisDualWriter


kishorekasi:
Add Jedis pipeline operations support for DynoJedisDualWriter client.

PR #243, opened 2018-10-19
Pipeline support in DynoDualWriterClient (#237)

kishorekasi:
Dyno DualWriter client to support all pipeline commands. All Redis commands sent to the shadow dyno client are async and only data mutation pipeline commands are sent to both primary and shadow dyno clients. All read pipeline commands are send to only primary dyno client.


ipapapa:
We can probably discuss it in a separate thread... In a future version, we can have the client wait until the delivery of the mutation successfully in both clusters (as you propose above) before returning back but this will increase the latency. However, if the latency is good then we can optionally add it if the application wants it. For example, a fast property to control whether we have it async or blocking.

kishorekasi:
I agree that adding fast properties will be a good way to control this behavior.

kishorekasi:
run travis-ci

PR #244, opened 2018-10-20
Redis authentication support

davidwadden:
Inspired by #187 

I'm interested in deploying Conductor to an environment that mandates Redis authentication.  I've followed a similar pattern as #187 for adding a `password` [field](https://github.com/Netflix/dyno/compare/master...davidwadden:redis-auth?expand=1#diff-8bf95e0cca7a2241e43ee9a278090157R50) to `Host.java` which then in turn is passed through the [`HostsUpdater.java`](https://github.com/Netflix/dyno/compare/master...davidwadden:redis-auth?expand=1#diff-623ee3b9f19280a0e5b584c1eba255d9) and set at the construction of the [`JedisConnectionFactory.java`](https://github.com/Netflix/dyno/compare/master...davidwadden:redis-auth?expand=1#diff-0acb68290b669a8b8e8ff2732284434a).

Interestings / questions:
- I only added a single [constructor overload](https://github.com/Netflix/dyno/compare/master...davidwadden:redis-auth?expand=1#diff-8bf95e0cca7a2241e43ee9a278090157R97) to `Host.java`.  Do you want any additional overloads that accept the `password`?.
- The [RedisAuthenticationIntegrationTest.java](https://github.com/Netflix/dyno/compare/master...davidwadden:redis-auth?expand=1#diff-8d372d34aa0e35a84da2b69b2d3d058e) that was added to cover Redis authentication introduces a `testCompile` dependency on [`embedded-redis`](https://github.com/Netflix/dyno/compare/master...davidwadden:redis-auth?expand=1#diff-c197962302397baf3a4cc36463dce5eaR90)
  - This dependency does not work on Windows, prefer not to have this test at all?
- `Host#toString` [masks](https://github.com/Netflix/dyno/compare/master...davidwadden:redis-auth?expand=1#diff-8bf95e0cca7a2241e43ee9a278090157R221) the contents of the `password` field.  Just a little non-standard
- Any additional test coverage I may have missed?

Also, noticed constructing a `Host` to be getting a bit complicated with the overloads.  Would you entertain another PR to introduce a `HostBuilder` to simplify this?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -87,6 +87,7 @@ project(':dyno-jedis') {
         compile project(':dyno-core')
         compile project(':dyno-contrib')
         compile 'redis.clients:jedis:2.9.0'
+        testCompile 'com.netflix.spinnaker.embedded-redis:embedded-redis:0.8.0'
     }
 }
 
--- a/dyno-jedis/src/test/java/com/netflix/dyno/jedis/RedisAuthenticationIntegrationTest.java
+++ b/dyno-jedis/src/test/java/com/netflix/dyno/jedis/RedisAuthenticationIntegrationTest.java
@@ -1,0 +1,245 @@
+package com.netflix.dyno.jedis;
+
+import com.google.common.base.Throwables;
+import com.netflix.dyno.connectionpool.Connection;
+import com.netflix.dyno.connectionpool.ConnectionPoolConfiguration;
+import com.netflix.dyno.connectionpool.Host;
+import com.netflix.dyno.connectionpool.Host.Status;
+import com.netflix.dyno.connectionpool.HostConnectionPool;
+import com.netflix.dyno.connectionpool.HostSupplier;
+import com.netflix.dyno.connectionpool.TokenMapSupplier;
+import com.netflix.dyno.connectionpool.exception.DynoConnectException;
+import com.netflix.dyno.connectionpool.impl.ConnectionPoolConfigurationImpl;
+import com.netflix.dyno.connectionpool.impl.CountingConnectionPoolMonitor;
+import com.netflix.dyno.connectionpool.impl.HostConnectionPoolImpl;
+import com.netflix.dyno.connectionpool.impl.lb.HostToken;
+import com.netflix.dyno.contrib.DynoOPMonitor;
+import java.net.ConnectException;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Assume;
+import org.junit.Before;
+import org.junit.Test;
+import redis.clients.jedis.Jedis;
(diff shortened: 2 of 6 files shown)


dyno-jedis/src/test/java/com/netflix/dyno/jedis/RedisAuthenticationIntegrationTest.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Test;
+import redis.embedded.RedisServer;
+import redis.embedded.RedisServerBuilder;
+
+public class RedisAuthenticationIntegrationTest {
```
kishorekasi: It will be useful to also add another testcase showing failure to connect to redisServer (with Auth) and client (no auth).

build.gradle
```diff
@@ -87,6 +87,7 @@ project(':dyno-jedis') {
         compile project(':dyno-core')
         compile project(':dyno-contrib')
         compile 'redis.clients:jedis:2.9.0'
+        testCompile 'com.netflix.spinnaker.embedded-redis:embedded-redis:0.8.0'
```
kishorekasi: If embedded-redis doesn't work on Windows, use conditional ignore (https://junit.org/junit4/javadoc/4.12/org/junit/Assume.html) to ignore such tests

kishorekasi: Also I am curious to know why embedded-redis doesn't work on windows

davidwadden: It looks like a change introduced on the Spinnaker [fork](https://github.com/spinnaker/embedded-redis) of [kstyrc/embedded-redis](https://github.com/kstyrc/embedded-redis) (which smelled abanoneded to me).  I saw it noted on the [README](https://github.com/spinnaker/embedded-redis#redis-version) of the Spinnaker fork.

The reason is the library actually embeds OS-specific executables that are executed as separate processes.

I will follow your recommendation to conditionally ignore this test and make sure things are green on Windows.

dyno-jedis/src/test/java/com/netflix/dyno/jedis/RedisAuthenticationIntegrationTest.java
davidwadden: I had to drop from the `DynoJedisClient` down to the `JedisConnectionFactory` to get any kind of valuable unhappy path tests.  I've added [test methods](https://github.com/davidwadden/dyno/blob/redis-auth/dyno-jedis/src/test/java/com/netflix/dyno/jedis/RedisAuthenticationIntegrationTest.java#L90-L207) to cover these cases from a lower layer, with the following cases added:

- Connect failed
- Password required
- Invalid password

PR #248, opened 2018-11-20
[WIP] Expire hash - Redis hash with ttl on fields.

kishorekasi:
[Sending this out for early feedback]

With expire hash, timeouts can be set on the hash fields. Expire hash uses sorted sets to store timeouts on the hash fields and will purge the fields on expiry.

ToDo:
- Add testcases


dyno-jedis/src/main/java/com/netflix/dyno/jedis/OpName.java
```diff
@@ -22,8 +22,9 @@
 	 DECR, DECRBY, DEL, DUMP, 
 	 ECHO, EXISTS, EXPIRE, EXPIREAT, EVAL,
 	 GET, GETBIT, GETRANGE, GETSET, 
-	 HDEL, HEXISTS,  HGET, HGETALL, HINCRBY, HINCRBYFLOAT, HKEYS, HLEN, HMGET, HMSET, HSET, HSCAN, HSETNX, HVALS, 
-	 INCR, INCRBY, INCRBYFLOAT, 
+	 HDEL, HEXISTS,  HGET, HGETALL, HINCRBY, HINCRBYFLOAT, HKEYS, HLEN, HMGET, HMSET, HSET, HSCAN, HSETNX, HVALS,
+	 EHSET, EHGET, EHDEL, EHEXISTS, EHGETALL, EHKEYS, EHMGET, EHVALS, EHMSET, EHSETNX, EHSCAN, EHINCRBY, EHINCRBYFLOAT,
```
shailesh33: Are these custom dynomite specific commands?

kishorekasi: These commands are dyno specific and build on top of existing Jedis commands.

dyno-demo/src/main/resources/demo.properties
```diff
@@ -4,6 +4,8 @@
 LOCAL_DATACENTER=us-east-1
```
smukil: Do we want to check this in as well? Or was this meant only for your local env?

kishorekasi: Oh! this was meant to be testing only. This file should not have any changes; I will revert them.

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
```diff
@@ -4060,6 +4066,530 @@ public long pfcount(String... keys) {
...
+        return hResponse.get();
+    }
+
+    @Override
+    public String ehget(final String key, final String field)
+            throws UnsupportedOperationException, DynoException {
+        final String dataKey = ehashDataKey(key);
+        final DynoJedisPipeline pipeline = this.pipelined();
+
+        EHMetadataUpdateResult ehMetadataUpdateResult = ehPurgeExpiredFields(pipeline, key);
```
smukil: It seems like a lot of functions have very similar steps. To avoid code duplication, is it possible to create a template and call that template once per function?

Eg: Have a more generic function like ehTask() or something:
ehTask(pipeline, key, field, func_ptr);

'func_ptr' would be 'hget', or 'hexists', etc.

And the last error checking step would be generic to all types of queries.

What do you think? This can be added as a future patch too.

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisClient.java
```diff
@@ -4060,6 +4066,529 @@ public long pfcount(String... keys) {
...
     }
 
+    private boolean validHashtag(final String hashtag) {
+        return !Strings.isNullOrEmpty(hashtag) && hashtag.length() == 2;
+    }
+
+    private String ehashDataKey(String key) throws UnsupportedOperationException {
+        String hashtag = connPool.getConfiguration().getHashtag();
+        if (!validHashtag(hashtag)) {
+            throw new UnsupportedOperationException("hashtags not set");
```
ipapapa: Why are we throwing an `UnsupportedOperationException` if the hashtag is not set and not another type of exception?

kishorekasi: `IllegalStateException` makes more sense instead.

PR #250, opened 2018-12-10
Fix issues in Dyno pipelines

kishorekasi:
-  Change Dyno dual writer pipeline to use single thread executor to
    run pipeline commands to avoid cases where some of the executor tasks may
    not run before the pipeline sync.

-  If dynomite cluster has hashkeys config set and if pipeline keys do not use
    hashtags, pipeline will end up using different connections for each
    pipeline command, which can result in unexpected outcomes and also result
    exhausting connections in connection pool. This is fixed by comparing
    the keys with the pipeline key when hashtags are not part of the key but
    dyno clusters have hashtags enabled.

### DIFF
--- a/dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoDualWriterPipeline.java
+++ b/dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoDualWriterPipeline.java
@@ -24,7 +24,7 @@ import java.util.concurrent.Future;
  */
 public class DynoDualWriterPipeline extends DynoJedisPipeline {
     private static final org.slf4j.Logger logger = LoggerFactory.getLogger(DynoDualWriterPipeline.class);
-    private static ExecutorService executor = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors());
+    private static ExecutorService executor = Executors.newSingleThreadExecutor();
     private final ConnectionPoolImpl<Jedis> connPool;
     private final DynoJedisPipeline shadowPipeline;
     private final DynoDualWriterClient.Dial dial;
--- a/dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisPipeline.java
+++ b/dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisPipeline.java
@@ -15,6 +15,7 @@
  */
 package com.netflix.dyno.jedis;
 
+import com.google.common.base.Strings;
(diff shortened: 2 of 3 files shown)


dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoDualWriterPipeline.java
```diff
@@ -24,7 +24,7 @@
  */
 public class DynoDualWriterPipeline extends DynoJedisPipeline {
     private static final org.slf4j.Logger logger = LoggerFactory.getLogger(DynoDualWriterPipeline.class);
-    private static ExecutorService executor = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors());
+    private static ExecutorService executor = Executors.newSingleThreadExecutor();
```
smukil: Any performance implications because of this? Or were we never supposed to use a ThreadPool in the first place?

Is it possible to "join" on the threads (or is there some mechanism to wait on the tasks running on the threads to complete) as a pre-condition to the sync() ?

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisPipeline.java
```diff
@@ -220,6 +221,9 @@ private void checkKey(final String key) {
 			 */
 			String hashValue = StringUtils.substringBetween(key, Character.toString(hashtag.charAt(0)),
 					Character.toString(hashtag.charAt(1)));
+			if (Strings.isNullOrEmpty(hashValue)) {
```
smukil: Shouldn't we rather just check if the pipeline is using hashtags at pipeline creation time and return an error saying that the Dynomite config has hashtags enabled, and therefore so should the pipeline itself?

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoDualWriterPipeline.java
kishorekasi: We were not supposed to use ThreadPool.

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisPipeline.java
kishorekasi: hashtags are set in the config and hashtags are checked during pipeline operations. So, this cannot be checked at pipeline creation time.

PR #267, opened 2019-04-18
Update topology with remote rack info when LOCAL_RACK and EC2_AVAILABILITY_ZONE are not specified

rsrinivasanNetflix:
When we do not set LOCAL_RACK and EC2_AVAILABILITY_ZONE we currently fall back to remote racks but do not update the topology to reflect the change.

### DIFF
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallback.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallback.java
@@ -1,12 +1,12 @@
 /*******************************************************************************
  * Copyright 2011 Netflix
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -53,7 +53,7 @@ import java.util.concurrent.atomic.AtomicReference;
  * outage in the local rack.
  * <p>
  * Note that this class does not prefer any one remote HostSelectionStrategy over another.
- *  
+ *
  * @author poberai
  * @author jcacciatore
  *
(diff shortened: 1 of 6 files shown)


dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallback.java
```diff
@@ -358,15 +358,18 @@ public void initWithHosts(Map<Host, HostConnectionPool<CL>> hPools) {
...
 
 		// Initialize Remote selectors
 		Set<String> remoteRacks = new HashSet<String>();
 		for (Host host : hPools.keySet()) {
 			String rack = host.getRack();
-			if (localRack != null && !localRack.isEmpty() && rack != null && !rack.isEmpty() && !localRack.equals(rack)) {
+			if(localRack == null && rack != null) {
+				remoteRacks.add(rack);
+			}
+			else if (localRack != null && !localRack.isEmpty() && rack != null && !rack.isEmpty() && !localRack.equals(rack)) {
```
smukil: Isn't the same as just removing the 'localrack != null' check in L372? That way we get the same behavior in one if condition.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallback.java
```diff
@@ -383,18 +386,22 @@ public void initWithHosts(Map<Host, HostConnectionPool<CL>> hPools) {
...
+		if (dataCenter == null) {
+		    if(localRack != null) {
+				dataCenter = localRack.substring(0, localRack.length() - 1);
+			}
+		}
 
-        for (HostToken hostToken: uniqueHostTokens) {
-        	if (hostToken.getHost().getRack().contains(dataCenter)) {
+		for (HostToken hostToken: uniqueHostTokens) {
+		    if(dataCenter == null || hostToken.getHost().getRack().contains(dataCenter)) {
```
smukil: This block of code seems like it only wants to obtain the hosts that belong in a specific datacenter. By adding the 'dataCenter == null' check, wouldn't that mean ALL datacenters are now added to the 'groups' map?

If that's correct, then it would probably make more sense to set a default DC as an 'else' condition to the 'if' statement in L398.

rsrinivasanNetflix:
> A high level philosophical comment I have is to add function comments to any existing functions you modify or new ones that you add, so that it'll be a lot easier to work with those same pieces of code the next time.
> 
> Not particularly necessary to make it a part of this patch, but it will be a good practice to have going forward for all our code bases.

Good point. I'll make those changes.

PR #269, opened 2019-05-24
Autoformatting code

rsrinivasanNetflix:
Autoformatting code. No functional code changes.


ipapapa:
I am not sure what is the best way. Autoformat all files at once or autoformat only the files that are being changed in each PR. I understand that this will handle everything at once but then it may add overhead when studying the commit history. Just my 2 cents.

rsrinivasanNetflix:
> I am not sure what is the best way. Autoformat all files at once or autoformat only the files that are being changed in each PR. I understand that this will handle everything at once but then it may add overhead when studying the commit history. Just my 2 cents.

Good point. We could do it that way. I like not polluting the commit history. Let me close this and add changes on future PRs to auto format.

rsrinivasanNetflix:
Merging this as is since the other route of incremental formatting makes future PRs difficult to read.

PR #270, opened 2019-06-18
Adding fix for host selection fallback using localrack

rsrinivasanNetflix:
Will add a test for this in a separate PR.


dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/lb/HostSelectionWithFallback.java
```diff
@@ -401,7 +401,8 @@ int calculateReplicationFactorForDC(List<HostToken> allHostTokens, String dataCe
                 dataCenter = localRack.substring(0, localRack.length() - 1);
             } else {
                 // No DC specified. Get the DC from the first host and use its replication factor
-                dataCenter = allHostTokens.get(0).getHost().getRack().substring(0, localRack.length() - 1);
```
smukil: LGTM, but just curious about how this was found. Does this happen when 'localRack' isn't set in the configuration?

PR #271, opened 2019-06-20
Implementation of Redlock

rsrinivasanNetflix:
Implementing redlock in Dyno.

Will add more tests. Starting PR to get some feedback to iterate.


dyno-jedis/src/main/java/com/netflix/dyno/jedis/operation/MultiKeyOperation.java
```diff
@@ -0,0 +1,57 @@
...
+        this.op = o;
+    }
+
+    @Override
+    public String getName() {
+        return op.name();
+    }
+
+    @Override
+    public String getStringKey() {
```
smukil: Would this be a source of confusion? Since we're sending back only the name of the first key? Maybe a comment about why this is the case would be useful.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/HostBuilder.java
```diff
@@ -0,0 +1,76 @@
...
+    public HostBuilder setDatastorePort(int datastorePort) {
+        this.datastorePort = datastorePort;
+        return this;
+    }
+
+    public Host createHost() {
+        if (datacenter == null) {
+            datacenter = ConfigUtils.getDataCenterFromRack(rack);
+        }
+        return new Host(hostname, ipAddress, port, securePort, datastorePort, rack, datacenter, status, hashtag, password);
```
smukil: Any pre-checks required? Or will the Host() constructor throw reliable error info on incorrect/incomplete use of the HostBuilder?

dyno-recipes/src/main/java/com/netflix/dyno/recipes/lock/DynoLockClient.java
```diff
@@ -0,0 +1,369 @@
...
+ */
+public class DynoLockClient {
+
+    private static final Logger logger = LoggerFactory.getLogger(DynoJedisClient.class);
+
+    private final ConnectionPool pool;
+    private final VotingHostsSelector votingHostsSelector;
+    private final ExecutorService service;
+    private final int quorum;
+    private final double CLOCK_DRIFT = 0.01;
```
smukil: A comment for this would be fairly important.

dyno-recipes/src/main/java/com/netflix/dyno/recipes/lock/DynoLockClient.java
```diff
@@ -0,0 +1,369 @@
...
+                );
+        boolean latchValue = false;
+        try {
+            latchValue = latch.await(timeout, timeoutUnit);
+        } catch (InterruptedException e) {
+            logger.info("Interrupted while releasing the lock for resource {}", resource);
+        }
+        if (latchValue) {
+            logger.info("Released lock on {}", resource);
+        } else {
```
smukil: `else` condition unnecessary. The above `catch` can relay the same information.

dyno-recipes/src/main/java/com/netflix/dyno/recipes/lock/DynoLockClient.java
```diff
@@ -0,0 +1,369 @@
...
+     */
+    private TimerTask getExtensionTask(Timer runJob, String resource, long ttlMS, Consumer<String> extensionFailed) {
+        return new TimerTask() {
+            @Override
+            public void run() {
+                long extendedValue = extendLock(resource, ttlMS);
+                if (extendedValue > 0) {
+                    logger.info("Extended lock on {} for {} MS", resource, ttlMS);
+                    TimerTask task = getExtensionTask(runJob, resource, ttlMS, extensionFailed);
+                    runJob.schedule(task, extendedValue/2);
```
smukil: A quick comment about how you decide when to reschedule here or in the function header comment would be useful.

dyno-recipes/src/main/java/com/netflix/dyno/recipes/lock/DynoLockClient.java
```diff
@@ -0,0 +1,369 @@
...
+        }
+        resourceKeyMap.remove(resource);
+    }
+
+    /**
+     * Return a timer task which will recursively schedule itself when extension is successful.
+     * @param runJob
+     * @param resource
+     * @param ttlMS
+     * @param extensionFailed
```
smukil: A quick comment would be good about this Consumer. (eg: called on failure)

dyno-recipes/src/main/java/com/netflix/dyno/recipes/lock/DynoLockClient.java
```diff
@@ -0,0 +1,369 @@
...
+            }
+        };
+    }
+
+    /**
+     * Try to acquire lock on resource for ttlMS and keep extending it by ttlMS when its about to expire.
+     * Calls the failure function with the resource when extension fails.
+     * @param resource
+     * @param ttlMS
+     * @param failure
```
smukil: ditto

dyno-recipes/src/main/java/com/netflix/dyno/recipes/lock/DynoLockClient.java
```diff
@@ -0,0 +1,369 @@
...
+            releaseLock(r);
+            failure.accept(r);
+        });
+    }
+
+    private boolean acquireLockWithExtension(String resource, long ttlMS, Consumer<String> extensionFailedCallback) {
+        long acquireResult = acquireLock(resource, ttlMS);
+        if (acquireResult > 0) {
+            Timer runJob = new Timer(resource, true);
+            runJob.schedule(getExtensionTask(runJob, resource, ttlMS, extensionFailedCallback), acquireResult/2);
```
smukil: `acquireResult/2`
Same comment as above.

dyno-recipes/src/main/java/com/netflix/dyno/recipes/lock/DynoLockClient.java
```diff
@@ -0,0 +1,369 @@
...
+        } else {
+            long startTime = Instant.now().toEpochMilli();
+            CopyOnWriteArrayList<Long> resultTtls = new CopyOnWriteArrayList<>();
+            CountDownLatch latch = new CountDownLatch(quorum);
+            votingHostsSelector.getVotingHosts().getEntireList().stream()
+                    .map(host -> new CheckAndRunHost(host, pool, "pttl", resource, resourceKeyMap.get(resource)))
+                    .forEach(checkAndRunHost -> CompletableFuture.supplyAsync(checkAndRunHost, service)
+                            .thenAccept(r -> {
+                                String result = r.getResult().toString();
+                                if (result.equals("0") || result.equals("-2")) {
```
smukil: Comment about why it's 0 and -2. (Can also set a local const variable to represent both the values if you like)

dyno-recipes/src/main/java/com/netflix/dyno/recipes/lock/DynoLockClient.java
```diff
@@ -0,0 +1,369 @@
...
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+
+/**
+ * Client for acquiring locks similar to the redlock implementation https://redis.io/topics/distlock
+ */
```
smukil: I think documenting some of the important implementation details here would be very helpful for users and debugging for developers.

The kind of details that would be useful are:
- The working of the common locking case
- What happens when one or more hosts are unreachable during acquire/extendLock()
- What happens when one or more hosts are unreachable during releaseLock()
- How we choose to reschedule extend() tasks (and why we choose to do it that way).
- Clock drift considerations.
- Possible edge case races we should be worried about.

dyno-jedis/src/main/java/com/netflix/dyno/jedis/DynoJedisUtils.java
```diff
@@ -0,0 +1,168 @@
...
+             */
+            if (hostTokens.isEmpty()) {
+                throw new DynoConnectException("No hosts in the TokenMapSupplier");
+            }
+        } else {
+            throw new DynoConnectException("TokenMapSupplier not provided");
+        }
+
+        String hashtag = hostTokens.get(0).getHost().getHashtag();
+        short numHosts = 0;
```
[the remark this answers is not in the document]
rsrinivasanNetflix: yup. that's correct. I'll refactor to this. I had just moved this code from the class to this Util.

dyno-jedis/src/main/java/com/netflix/dyno/jedis/operation/MultiKeyOperation.java
rsrinivasanNetflix: For some context, this was a refactor where I pulled these classes out from DynoJedisClient since they're useful outside also. Will add some comments.

dyno-recipes/src/main/java/com/netflix/dyno/recipes/lock/DynoLockClient.java
rsrinivasanNetflix: The catch is if the thread gets interrupted. The else is when the latch hits a time out in which case we return immediately.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/HostBuilder.java
rsrinivasanNetflix: preserving status quo with the old implementation. There might be some cases where we should be throwing an error but arent. we'd need to look into that in a separate PR.

dyno-recipes/src/main/java/com/netflix/dyno/recipes/lock/DynoLockClient.java
rsrinivasanNetflix: I've added some comments on the major points. Will add a few more in the wiki for this. Working on the wiki right now.

dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostsUpdater.java
```diff
@@ -107,28 +114,27 @@ public HostStatusTracker refreshHosts() {
             if (hostFromHostSupplier.isUp()) {
                 Host hostFromTokenMapSupplier = allHostSetFromTokenMapSupplier.get(hostFromHostSupplier);
 
-                hostsUpFromHostSupplier.add(new Host(hostFromHostSupplier.getHostName(), hostFromHostSupplier.getIpAddress(),
-                        hostFromTokenMapSupplier.getPort(), hostFromTokenMapSupplier.getSecurePort(), hostFromTokenMapSupplier.getRack(),
-                        hostFromTokenMapSupplier.getDatacenter(), Host.Status.Up, hostFromTokenMapSupplier.getHashtag(),
-                        hostFromTokenMapSupplier.getPassword()));
+                hostsUpFromHostSupplier.add(new HostBuilder().setHostname(hostFromHostSupplier.getHostName())
```
jcacciatore: nit - add a convenience method that creates a Host from a given Host to avoid duplicating the code to copy it over here and both places below

rsrinivasanNetflix: good point. updated!

dyno-recipes/src/main/java/com/netflix/dyno/recipes/lock/VotingHostsFromTokenRange.java
```diff
@@ -0,0 +1,90 @@
...
+            List<HostToken> allHostTokens = tokenMapSupplier.getTokens(ImmutableSet.copyOf(hostSupplier.getHosts()));
+            if (allHostTokens.size() < MIN_VOTING_SIZE) {
+                throw new IllegalStateException(String.format("Cannot perform voting with less than %d nodes", MIN_VOTING_SIZE));
+            }
+            // Total number of hosts present per rack
+            Map<String, Long> numHostsPerRack = allHostTokens.stream().map(ht -> ht.getHost().getRack()).collect(Collectors.groupingBy(Function.identity(), Collectors.counting()));
+            AtomicInteger numHostsRequired = new AtomicInteger(effectiveVotingSize);
+            // Map to keep track of number of hosts to take for voting from this rack
+            Map<String, Integer> numHosts = new HashMap<>();
+            // Sort racks to get the same order
```
smukil: Would this work reliably if we have the same rack names across DCs?

dyno-recipes/src/main/java/com/netflix/dyno/recipes/lock/VotingHostsFromTokenRange.java
```diff
@@ -0,0 +1,90 @@
...
+            List<String> racks = numHostsPerRack.keySet().stream().sorted(Comparator.comparing(String::toString)).collect(Collectors.toList());
+            for(String rack: racks) {
+//                // Already have required number of hosts for voting. Do not take any more.
+//                if(numHostsRequired.get() <= 0) {
+//                    numHosts.put(rack, 0);
+//                    continue;
+//                }
+                // Need more hosts take as as many as you can from this rack.
+                int v = (int) Math.min(numHostsRequired.get(), numHostsPerRack.get(rack));
```
smukil: How do we handle the case where we don't have enough voting hosts? Eg: If we just get back one host from this?

dyno-recipes/src/main/java/com/netflix/dyno/recipes/lock/VotingHostsFromTokenRange.java
```diff
@@ -0,0 +1,90 @@
...
+            // Total number of hosts present per rack
+            Map<String, Long> numHostsPerRack = allHostTokens.stream().map(ht -> ht.getHost().getRack()).collect(Collectors.groupingBy(Function.identity(), Collectors.counting()));
+            AtomicInteger numHostsRequired = new AtomicInteger(effectiveVotingSize);
+            // Map to keep track of number of hosts to take for voting from this rack
+            Map<String, Integer> numHosts = new HashMap<>();
+            // Sort racks to get the same order
+            List<String> racks = numHostsPerRack.keySet().stream().sorted(Comparator.comparing(String::toString)).collect(Collectors.toList());
+            for(String rack: racks) {
+//                // Already have required number of hosts for voting. Do not take any more.
+//                if(numHostsRequired.get() <= 0) {
```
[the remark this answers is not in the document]
rsrinivasanNetflix: good point. this got left behind. will remove.

dyno-recipes/src/main/java/com/netflix/dyno/recipes/lock/VotingHostsFromTokenRange.java
rsrinivasanNetflix: good point. added a check for an edge case.

dyno-recipes/src/main/java/com/netflix/dyno/recipes/lock/VotingHostsFromTokenRange.java
rsrinivasanNetflix: resolved offline

Issue #273, opened 2019-07-04
NullPointerException when connecting to multiple hosts


emiteze:
I'm currently testing some behaviors on a local dynomite cluster using Docker and Docker Compose and I'm getting a NullPointer when I try to connect to multiple hosts. When using multiple hosts like this `dynomite.cluster.hosts=localhost:8102:dc1-rack1:dc1:2147483647,localhost:8103:dc1-rack1:dc1:4294967294,localhost:8104:dc1-rack2:dc1:4294967295` I get this error when running my test (just test a insertion on dynomite).

```
Caused by: java.lang.NullPointerException
	at com.netflix.dyno.connectionpool.impl.HostsUpdater.refreshHosts(HostsUpdater.java:111)
```

The NullPointer comes from this point where hostFromTokenMapSupplier is Null.

https://github.com/Netflix/dyno/blob/8af6409746e3d10b15c9e5c5d458dcbf64537478/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostsUpdater.java#L110

@chrisbendel you said on #264 that you had this issue, how do you managed to solve it?

Anyone else has some tips about this?

Thanks in advance :)

chrisbendel:
@emiteze It's been a while since I spent some time using this, but from what I remember my workaround was like I said in the issue was clearing the hosts. We've been using it in production since then and no issues have come up.

I'd be interested to get back into it and try debugging it when I have some free time at work to spend on the issue.

If you figure out any new info for debugging purposes, definitely let me know here in this issue thread. I'd be curious what any maintainers might say if they happen to check on the issues here. I'm sure it's something you and I are doing wrong configuration-wise, but you never know if there's a bug somewhere in the source

Issue #264, opened 2019-04-12
Duplicated hosts in custom HostSupplier implementation


chrisbendel:
Observed weird/confusing behavior when implementing a dyno client into our Spring application.  
Dynoclient is being initiated as a singleton service (using spring @Service annotation). This could all very well be my misunderstanding of how spring's service annotation works under the hood, or just bad code in general. Looking to get a bit more information and understanding of what might be going on. I am suspicious that getHosts() is being called twice during the initialization of the client, and the hosts variable inside customHostSupplier is having the two hosts added twice.  
  
Basic information:  
- Two hosts on two different racks, same datacenter  
- Before clearing the hosts (see the // hack comment in code below), we were hitting unexpected NullPointerExceptions [around here (hostFromTokenMapSupplier was null)](https://github.com/Netflix/dyno/blob/58d88e2dd53ff430f200e096be86e5a1419ea3c8/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostsUpdater.java#L110)  
- We essentially had 4 hosts (one dupe of each actual host), so clearing out the hosts in my customHostSupplier forced it to have the correct number of hosts  
  
Hoping to maybe gain some more knowledge on things, for all I know I could be doing something wrong. Let me know if you need any more information.  
  
Disclaimer: Configurations seem to be all working and set up properly, things work with the hack in place

```
@Service
public class Dynosaur {
    private DynoJedisClient client;
    private static final Gson gson = new Gson();
    private Integer ttl;
    private Boolean enabled = true;

    public Dynosaur(
        // Params are coming from spring application.properties
        @Value("${dyno.servers}") List<String> servers,
        @Value("${dyno.racks}") List<String> racks,
        @Value("${dyno.port}") Integer port,
        @Value("${dyno.ttl}") Integer ttl,
        @Value("${dyno.token}") Long token,
        @Value("${dyno.enabled}") Boolean enabled
    ) {
        this.enabled = enabled;
        this.ttl = ttl;

        HostSupplier customHostSupplier = new HostSupplier() {
            private final List<Host> hosts = new ArrayList<>();

            @Override
            public List<Host> getHosts() {
                // The current hack to work around this
                hosts.clear();
                Integer i = 0;
                for (String server : servers) {
                    hosts.add(new Host(server, port, racks.get(i), Host.Status.Up));
                    i++;
                }
                return hosts;
            }
        };

        TokenMapSupplier tokenMapSupplier = new TokenMapSupplier() {
[...]

PR #281, opened 2019-08-30
Use IPAddresses from HostSupplier rather than from AbstractTokenMapSu…

smukil:
…pplier

We always used the IPs from the HostSupplier rather than the
AbstractTokenMapSupplier as an undocumented subtlety.

In the previous patch, this was changed as it seemed like a harmless
change.
https://github.com/Netflix/dyno/commit/7b253925f8bae6355a16164efe8469d7929f0591

Howver, the implication is that the client operates on Public IPs (which
the AbstractTokenMapSupplier supplies), as opposed to the Private IPs
(which the HostSupplier supplies), and in production, the Public IP
range are generally not accessible by clients.

Why does the AbstractTokenMapSupplier return public IPs?
This is because the AbstractTokenMapSupplier derives its values from
dynomite-manager which returns the topology seen by it and the server
processes, and they use Public IPs to communicate with each other.

TODO: Can we ensure that both the HostSupplier and AbstractTokenMapSupplier
agree on the same values?

### DIFF
--- a/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostsUpdater.java
+++ b/dyno-core/src/main/java/com/netflix/dyno/connectionpool/impl/HostsUpdater.java
@@ -16,6 +16,7 @@
 package com.netflix.dyno.connectionpool.impl;
 
 import com.netflix.dyno.connectionpool.Host;
+import com.netflix.dyno.connectionpool.HostBuilder;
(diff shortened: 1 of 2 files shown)


rsrinivasanNetflix:
thanks for fixing this @smukil !

chrisbendel:
Could this have possibly fixed: #273 and #264  ? Would be good to know to close out some issues.

PR #288, opened 2020-01-22
exclude lombok from the compile classpath

joshgord:
By moving to compileOnly this will prevent lombok from being defined as a transitive dependency on the client

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -98,7 +98,7 @@ project(':dyno-jedis') {
         compile project(':dyno-core')
         compile project(':dyno-contrib')
         compile 'redis.clients:jedis:3.0.1'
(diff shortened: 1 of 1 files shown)


smukil:
LGTM. Adding @akhaku as well.

build.gradle
```diff
@@ -146,6 +146,7 @@ project(':dyno-recipes') {
         compile project(':dyno-core')
         compile project(':dyno-jedis')
         compile 'com.google.code.gson:gson:2.8.5'
+        compile 'org.projectlombok:lombok:1.18.4'
```
akhaku: curious: why is it needed in the `compile` configuration here?

PR #310, opened 2021-03-18
Replace JCenter with Maven Central

rpalcolea:
This replaces jcenter with maven central to avoid broken builds in the future


rpalcolea:
This should be ready for review @smukil

